Issue #6, opened 2022-08-14
Add IPC support for host allocations


jandres742:
Allow for USM host allocations to be also exported as IPC handles, as it is defined currently for device allocations 

https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=ipchandle#zememgetipchandle

> Creates an IPC memory handle for the specified allocation.
>
>Takes a pointer to a device memory allocation and creates an IPC memory handle for exporting it for use in another process.
>
>The pointer must be base pointer of the device memory allocation; i.e. the value returned from [zeMemAllocDevice]>>(https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=ipchandle#ze__api_8h_1ad47b7ece6b6c412dbb96fbe0afd89147).

I dont think we need changes in the interface, as all it takes is a pointer. We just need to add this to the documentation.

bashbaug:
If we add IPC support for host allocations then we should support host memory for external memory import and export as well.

jandres742:
@bashbaug . @zackwaters : Please see here https://github.com/oneapi-src/level-zero-spec/pull/18

PR #18, opened 2022-09-30
Add host support for IPC allocations

jandres742:
Resolves: #6

### DIFF
--- a/scripts/core/PROG.rst
+++ b/scripts/core/PROG.rst
@@ -611,7 +611,11 @@ External memory handles may be imported from other APIs, or exported for use in 
 Importing and exporting external memory is an optional feature.
 Devices may describe the types of external memory handles they support using ${x}DeviceGetExternalMemoryProperties.
 
+%if ver >= 1.5:
+Importing and exporting external memory is supported for device and host memory allocations and images.
+%else:
 Importing and exporting external memory is supported for device memory allocations and images.
+%endif
 
 The following pseudo-code demonstrates how to allocate and export an external memory handle for a device memory allocation as a Linux dma_buf:
 
--- a/scripts/core/memory.yml
+++ b/scripts/core/memory.yml
@@ -290,7 +290,8 @@ name: GetIpcHandle
 decl: static
 details:
     - "Takes a pointer to a device memory allocation and creates an IPC memory handle for exporting it for use in another process."
-    - "The pointer must be base pointer of the device memory allocation; i.e. the value returned from $xMemAllocDevice."
+    - "1.0": "The pointer must be base pointer of a device memory allocation; i.e. the value returned from $xMemAllocDevice."
+      "1.5": "The pointer must be base pointer of a device or host memory allocation; i.e. the value returned from $xMemAllocDevice or from $xMemAllocHost, respectively."
     - "The application may call this function from simultaneous threads."
     - "The implementation of this function must be thread-safe."
 params:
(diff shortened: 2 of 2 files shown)


bashbaug (changes requested):
I think we'll want a few more changes:

1. For ze_external_memory_export_desc_t and ze_external_memory_import_*_t we need to update the descriptions to say that they can be passed to zeMemAllocHost as well as zeMemAllocDevice.
2. We currently don't differentiate between device allocations and host allocations in zeDeviceGetExternalMemoryProperties().  Is this OK, or do we ever expect to support some types of external memory for device allocations and some other types of external memory for host allocations?  If we think there might be a difference we should define a new version of ze_device_external_memory_properties_t for host allocations.
3. Finally, I think as-written these changes will be retroactive to all spec versions.  Since I think we want these changes to be v1.5-specific we should add version tags so earlier specifications are unchanged.

jandres742:
@bashbaug 

> do we ever expect to support some types of external memory for device allocations and some other types of external memory for host allocations?

I think we should have a differentiation, correct. To not limit us to the same types. I think we could have a new `zeDriverGetExternalMemoryProperties(ze_driver_external_memory_properties_t)`, and in there we would have all types:

```
ze_external_memory_type_flags_t hostMemoryAllocationImportTypes
[out] Supported external memory import types for host memory allocations.

ze_external_memory_type_flags_t hostMemoryAllocationExportTypes
[out] Supported external memory export types for host memory allocations.

ze_external_memory_type_flags_t deviceMemoryAllocationImportTypes
[out] Supported external memory import types for device memory allocations.

ze_external_memory_type_flags_t deviceMemoryAllocationExportTypes
[out] Supported external memory export types for device memory allocations.

// import images
// export images
```

we could even add shared memory allocation flags. If an implementation doesn't support for a given type, just doesn't set any flags.

jandres742:
@bashbaug 

> For ze_external_memory_export_desc_t and ze_external_memory_import_*_t we need to update the descriptions to say that they can be passed to zeMemAllocHost as well as zeMemAllocDevice.

Done

> We currently don't differentiate between device allocations and host allocations in zeDeviceGetExternalMemoryProperties(). Is this OK, or do we ever expect to support some types of external memory for device allocations and some other types of external memory for host allocations?

I would say this should be true. We can see in the future if that's not the case.

> Finally, I think as-written these changes will be retroactive to all spec versions. Since I think we want these changes to be v1.5-specific we should add version tags so earlier specifications are unchanged.

done

This needs to be merged after https://github.com/oneapi-src/level-zero-spec/pull/25

jandres742:
Rebased.

scripts/core/PROG.rst
```diff
@@ -611,7 +611,7 @@ External memory handles may be imported from other APIs, or exported for use in
 Importing and exporting external memory is an optional feature.
 Devices may describe the types of external memory handles they support using ${x}DeviceGetExternalMemoryProperties.
 
-Importing and exporting external memory is supported for device memory allocations and images.
+Importing and exporting external memory is supported for device and host memory allocations and images.
```
bashbaug, line 614:
```suggestion
%if ver >= 1.5:
Importing and exporting external memory is supported for device and host memory allocations and images.
%else:
Importing and exporting external memory is supported for device memory allocations and images.
%endif
```

I think we'll want version checks here because we only supported device allocations pre-v1.5.

scripts/core/memory.yml
```diff
@@ -396,7 +398,8 @@ members:
       name: fd
       desc: "[in] the file descriptor handle to import"
 details:
-    - "This structure may be passed to $xMemAllocDevice, via the `pNext` member of $x_device_mem_alloc_desc_t, to import memory from a file descriptor."
+    - "1.0": "This structure may be passed to $xMemAllocDevice, via the `pNext` member of $x_device_mem_alloc_desc_t, to import memory from a file descriptor."
+      "1.5": "This structure may be passed to $xMemAllocDevice or $xMemAllocHost, via the `pNext` member of $x_device_mem_alloc_desc_t or of $x_host_mem_alloc_desc_t, respectively, to import memory from a file descriptor."
```
bashbaug, line 402: This looks good but I think we need a similar change to `$x_external_memory_import_win32_handle_t` below?

bashbaug (commented):
This mostly looks good but I think we still need a few small fixes.

Also, just to confirm, we're OK postponing this until later?

> We currently don't differentiate between device allocations and host allocations in zeDeviceGetExternalMemoryProperties(). Is this OK, or do we ever expect to support some types of external memory for device allocations and some other types of external memory for host allocations? If we think there might be a difference we should define a new version of ze_device_external_memory_properties_t for host allocations.

scripts/core/PROG.rst
jandres742: @bashbaug  you are correct. Added.

scripts/core/memory.yml
jandres742: @bashbaug missed it. Added.

bashbaug (approved):
Thanks, LGTM!

PR #20, opened 2022-10-03
clarify that a context can also be used by sub-devices of devices

### LINKED ISSUE #8: Clarify that a context created against a root device includes automatically all its sub-devices
From @bashbaug:

> ... the proposal is that a context with a parent device will automatically work with its sub-devices.

So all that we need is to add that clarification in the spec. This is already an assumption driver was having. We just need clarification in the spec.

bashbaug:
fixes #8 

Clarifies that a context can be used by all of the devices and any sub-devices of the devices that were specified when the context was created.

Signed-off-by: Ben Ashbaugh <ben.ashbaugh@intel.com>

### DIFF
--- a/scripts/core/context.yml
+++ b/scripts/core/context.yml
@@ -81,8 +81,10 @@ params:
       name: phDevices
       desc: |
             [in][optional][range(0, numDevices)] array of device handles which context has visibility.
-            if nullptr, then all devices supported by the driver instance are visible to the context.
-            otherwise, context only has visibility to devices in this array.
+            if nullptr, then all devices and any sub-devices supported by the driver instance are
+            visible to the context.
+            otherwise, the context only has visibility to the devices and any sub-devices of the
+            devices in this array.
     - type: $x_context_handle_t*
       name: phContext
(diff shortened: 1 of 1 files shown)


scripts/core/context.yml
```diff
@@ -81,8 +81,10 @@ params:
       name: phDevices
       desc: |
             [in][optional][range(0, numDevices)] array of device handles which context has visibility.
-            if nullptr, then all devices supported by the driver instance are visible to the context.
-            otherwise, context only has visibility to devices in this array.
+            if nullptr, then all devices and any sub-devices supported by the driver instance are
+            visible to the context.
+            otherwise, the context only has visibility to the devices and any sub-devices of the
```
jandres742, line 86: nit: Upper case in O in `otherwise`.

bashbaug: The capitalization is the same as it was before this PR.  If we're going to change capitalization I'd prefer to change it uniformly and everywhere.

jandres742: we can do it in other PR.

PR #23, opened 2022-10-06
add an extension to return the device IP version

### LINKED ISSUE #15: add an extension to query device IP version
We need to add a query for the "device IP version".  This would be a simple addition to zeDeviceGetProperties, and could be added in a similar way as the device LUID extension.

Is there anything else we should add to the same extension while we're at it?

See: https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_device_attribute_query.html

Notes:
* We already have the device ID via ze_device_properties_t.deviceId.
* We already have the number of slices via ze_device_properties_t.numSlices.
* We already have the number of sub-slices per slice via ze_device_properties_t.numSubslicesPerSlice.
* We already have the number of EUs per sub-slice via ze_device_properties_t.numEUsPerSubslice.
* We already have the number of threads per EU via ze_device_properties_t.numThreadsPerEU.
* We already have the dp4a feature flag via ze_device_module_flags_t.ZE_DEVICE_MODULE_FLAG_DP4A.
* We don't appear to have a feature flag for dpas.

bashbaug:
Fixes #15

Signed-off-by: Ben Ashbaugh <ben.ashbaugh@intel.com>

### DIFF
--- a/scripts/core/common.yml
+++ b/scripts/core/common.yml
@@ -407,6 +407,9 @@ etors:
     - name: DEVICE_MEMORY_EXT_PROPERTIES
       desc: $x_device_memory_ext_properties_t
       version: "1.4"
+    - name: DEVICE_IP_VERSION_EXT
+      desc: $x_device_ip_version_ext_t
+      version: "1.5"
     - name: RELAXED_ALLOCATION_LIMITS_EXP_DESC
       desc: $x_relaxed_allocation_limits_exp_desc_t
       value: "0x00020001"
--- a/scripts/core/EXT_DeviceIpVersion.rst
+++ b/scripts/core/EXT_DeviceIpVersion.rst
@@ -1,0 +1,22 @@
+<%
+import re
+from templates import helper as th
+%><%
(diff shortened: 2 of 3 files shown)


scripts/core/common.yml
```diff
@@ -407,6 +407,9 @@ etors:
     - name: DEVICE_MEMORY_EXT_PROPERTIES
       desc: $x_device_memory_ext_properties_t
       version: "1.4"
+    - name: DEVICE_IP_VERSION_EXT
```
jandres742, line 410: should the name be EXT_PROPERTIES as with others?

jandres742: no, we are ok.

bashbaug: I considered this, but this query isn't returning any "properties", so I didn't think this made sense.

Note that the EU count query a few lines above also does not include "properties" in the name.

jandres742: right, I noticed the EU count as well. cool.

jandres742 (approved):

PR #25, opened 2022-10-07
Mark zesPowerGetLimits and zesPowerSetLimits as deprecated

### LINKED ISSUE #12: Remove or comment the deprecated APIs zesPowerGetLimits && zesPowerSetLimits
In the spec, new APIs zesPowerGetLimitsExt && zesPowerSetLimitsExt introduced but why we are still maintaining the old APIS
zesPowerGetLimits
zesPowerSetLimits

https://spec.oneapi.io/level-zero/latest/sysman/api.html#zespowergetlimits

it is better to remove them or comment in the API's saying deprecated and use new APIs

jandres742:
Resolves #12

### DIFF
--- a/scripts/sysman/power.yml
+++ b/scripts/sysman/power.yml
@@ -257,6 +257,7 @@ name: GetLimits
 details:
     - "The application may call this function from simultaneous threads."
     - "The implementation of this function should be lock-free."
+    - "1.5": "Note: This function is deprecated and replaced by $sPowerGetLimitsExt."
 params:
     - type: $s_pwr_handle_t
       name: hPower
@@ -278,6 +279,7 @@ name: SetLimits
 details:
     - "The application may call this function from simultaneous threads."
     - "The implementation of this function should be lock-free."
+    - "1.5": "Note: This function is deprecated and replaced by $sPowerSetLimitsExt."
 params:
     - type: $s_pwr_handle_t
       name: hPower
--- a/scripts/sysman/powerLimits.yml
+++ b/scripts/sysman/powerLimits.yml
@@ -85,7 +85,7 @@ name: GetLimitsExt
 details:
     - "The application may call this function from simultaneous threads."
     - "The implementation of this function should be lock-free."
-    - "This function returns all the power limits assocaited with the supplied power domain."
+    - "This function returns all the power limits associated with the supplied power domain."
 params:
     - type: $s_pwr_handle_t
       name: hPower
@@ -102,9 +102,9 @@ desc: "Set power limits"
 class: $sPower
 name: SetLimitsExt
 details:
-    - "The application can only modify unlocked members of the limit descriptors returned by ${s}PowerGetLimitsExt."
-    - "Not all the limits returned by ${s}PowerGetLimitsExt need to be supplied to this function."
-    - "Limits do not have to be supplied in the same order as returned by ${s}PowerGetLimitsExt."
+    - "The application can only modify unlocked members of the limit descriptors returned by $sPowerGetLimitsExt."
+    - "Not all the limits returned by $sPowerGetLimitsExt need to be supplied to this function."
+    - "Limits do not have to be supplied in the same order as returned by $sPowerGetLimitsExt."
     - "The same limit can be supplied multiple times. Limits are applied in the order in which they are supplied."
     - "The application may call this function from simultaneous threads."
     - "The implementation of this function should be lock-free."


jandres742:
@bashbaug please see

scripts/sysman/power.yml
```diff
@@ -257,6 +257,7 @@ name: GetLimits
 details:
     - "The application may call this function from simultaneous threads."
     - "The implementation of this function should be lock-free."
+    - "Note: This function is being deprecated and replaced by $sPowerGetLimitsExt."
```
bashbaug, line 260:
```suggestion
    - "Note: This function is deprecated and replaced by $sPowerGetLimitsExt."
```

I think it is more correct to say that it **is** deprecated, unless we really want to say that it it is **being** deprecated (at some point in the future, in other words not deprecated just yet)?

scripts/sysman/power.yml
```diff
@@ -278,6 +279,7 @@ name: SetLimits
 details:
     - "The application may call this function from simultaneous threads."
     - "The implementation of this function should be lock-free."
+    - "Note: This function is being deprecated and replaced by $sPowerSetLimitsExt."
```
bashbaug, line 282:
```suggestion
    - "Note: This function is deprecated and replaced by $sPowerSetLimitsExt."
```

scripts/sysman/power.yml
jandres742: Agree with your suggestion @bashbaug . Changing.

bashbaug: One more thought: this should be guarded by version 1.5 also, correct?  Otherwise this will apply retroactively to all spec versions.

jandres742: I was thinking on that. so something like

```cpp
- "1.5": "Note: This function is deprecated and replaced by $sPowerGetLimitsExt."
```

jandres742: should we mark the previous two lines with 1.0, or just marking this one with 1.5 is enough?

bashbaug (commented):
I think I'd rather use a YAML syntax similar to the descriptions (`desc`) rather than the syntax proposed here.  This syntax would look more like:

```yaml
details: 
    - "1.0": "This detail is valid in version 1.0."
      "1.5": "This detail is valid in version 1.5."
      "1.6": "This detail would be valid in version 1.6."
    - "1.0": "This is another detail that is valid in version 1.0."
      "1.2": "This is another detail that is valid in version 1.2."
```

This could be done by adding the following to parse_specs.py, in `__filter_version`:

```python
    def __filter_desc(d):
        if 'desc' in d and isinstance(d['desc'], dict):
            for k, v in d['desc'].items():
                if float(k) <= max_ver:
                    desc = v
            d['desc'] = desc
        return d

    # Start of new code:
    def __filter_detail(det):
        if isinstance(det, dict):
            detail = None
            for k, v in det.items():
                try:
                    version = float(k)
                except:
                    continue
                if version <= max_ver:
                    detail = v
            if detail:
                return detail
        return det

    if 'details' in d:
        flt = []
        for det in d['details']:
            flt.append(__filter_detail(det))
        d['details'] = flt
    # End of new code.

    flt = []
```

What do you think?

Note that this is a little more complicated than some other version checks because it needs to handle existing cases where dictionary descriptions are used to represent nested lists.  Here is an example that uses a nested list, for "The following are conditions for...":

```yaml
details:
    - "Any device can access any other device within a node through a scale-up fabric."
    - "The following are conditions for CanAccessPeer query.":
        - "If both device and peer device are the same then return true."
        - "If both sub-device and peer sub-device are the same then return true."
        - "If both are sub-devices and share the same parent device then return true."
        - "If both device and remote device are connected by a direct or indirect scale-up fabric or over PCIe (same root complex or shared PCIe switch) then true."
        - "If both sub-device and remote parent device (and vice-versa) are connected by a direct or indirect scale-up fabric or over PCIe (same root complex or shared PCIe switch) then true."
```

jandres742:
thanks @bashbaug . Changed as suggested, with minor changes to the proposed code for filter_version to make it work.

bashbaug:
What issues did you find in the proposed code for filter_version?  The snippet I sent worked for me:

https://github.com/bashbaug/level-zero-spec/tree/version-specific-details

jandres742:
Rebase on top of @bashbaug changes.

bashbaug (approved):

PR #27, opened 2022-10-12
Add sub-allocations properties extensions

jandres742:
Depending on target platform, certain Level Zero drivers may split an L0 allocation into smaller sub-allocations for performance purposes. This extension allows applications to query for the number and size of those sub-allocations.

### DIFF
--- a/scripts/core/EXT_Exp_SubAllocationProperties.rst
+++ b/scripts/core/EXT_Exp_SubAllocationProperties.rst
@@ -1,0 +1,70 @@
+<%
+import re
+from templates import helper as th
+%><%
+    OneApi=tags['$OneApi']
+    x=tags['$x']
+    X=x.upper()
+%>
+:orphan:
+
+.. _ZE_experimental_sub_allocations:
+
+=====================================
+ Sub-Allocation Properties Extension
+=====================================
+
+API
+----
+
+* Enumerations
+
+
+    * ${x}_sub_allocations_exp_version_t
+
+
+* Structures
+
+
+    * ${x}_memory_sub_allocations_exp_properties_t
+
+Sub-Allocation Properties
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Driver implementations may choose to create a device allocation as a series of sub-allocations. For instance,
+an allocation created against a parent device may be allocated internally as set of N sub-allocations, with
+N being the number of sub-devices associated with the parent device.
+
+The sub-allocation properties extension may be used to get the properties, i.e. base address and size, for
+each of those sub-allocations. The following pseudo-code demonstrates a basic use-case of this extension:
+
+.. parsed-literal::
+
+    ${x}MemAllocDevice(context, &desc, size, alignment, device, &ptr);
+
+    ${x}_memory_sub_allocations_exp_properties_t subAllocationDesc {};
+    uitn32_t numberOfSuballocations = 0;
+    subAllocationDesc.stype = ${X}_STRUCTURE_TYPE_MEMORY_SUB_ALLOCATIONS_EXP_PROPERTIES;
+    subAllocationDesc.pCount = &numberOfSuballocations;
+
+    ${x}_memory_allocation_properties_t memAllocProperties {};
+    memAllocProperties.stype = ${X}_STRUCTURE_TYPE_MEMORY_ALLOCATION_PROPERTIES;
+    memAllocProperties.pNext = &subAllocationDesc;
+
+    // Get number of sub-allocations
+    ${x}MemGetAllocProperties(context, ptr, &memAllocProperties, nullptr);
+
+    // if more than 1 sub-allocation, then allocation has been split
+    if (numberOfSuballocations > 1) {
+        std::vector<${x}_sub_allocation_t> subAllocationMemAllocProperties(numberOfSuballocations);
+        subAllocationDesc.pSubAllocations = subAllocationMemAllocProperties.data();
+
+        ${x}MemGetAllocProperties(context, ptr, &memAllocProperties, nullptr);
+
+        // retrieve the properties of each sub-allocation
+        for (auto &subAllocationProperty : subAllocationMemAllocProperties) {
+            void * base = subAllocationProperty.base;
+            size_t size = subAllocationProperty.size;
+        }
+    }
+    ...
--- a/scripts/core/common.yml
+++ b/scripts/core/common.yml
@@ -458,6 +458,10 @@ etors:
       desc: $x_fabric_edge_exp_properties_t 
       version: "1.4"
       value: "0x0002000C"
+    - name: MEMORY_SUB_ALLOCATIONS_EXP_PROPERTIES
+      desc: $x_memory_sub_allocations_exp_properties_t
+      version: "1.5"
+      value: "0x0002000D"
     - name: COMMAND_GRAPH_EXP_DESC
       desc: $x_command_graph_exp_desc_t
       version: "2.0"
--- a/scripts/core/subAllocationsProperties.yml
+++ b/scripts/core/subAllocationsProperties.yml
@@ -1,0 +1,60 @@
+#
+# Copyright (C) 2022 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+#
+# See YaML.md for syntax definition
+#
+--- #--------------------------------------------------------------------------
+type: header
+desc: "Intel $OneApi Level-Zero Extension for querying sub-allocations properties."
+version: "1.5"
+--- #--------------------------------------------------------------------------
+type: macro
+desc: "Sub-Allocations Properties Extension Name"
+version: "1.5"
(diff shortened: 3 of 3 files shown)


jandres742:
@bashbaug : Please see here.

scripts/core/common.yml
```diff
@@ -410,6 +410,9 @@ etors:
     - name: DEVICE_IP_VERSION_EXT
       desc: $x_device_ip_version_ext_t
       version: "1.5"
+    - name: SUB_ALLOCATIONS_EXT_PROPERTIES
+      desc: $x_sub_allocations_ext_properties_t
+      version: "1.5"
```
bashbaug, line 415: Since this is currently an experimental extension I think we only need the "EXP version below and this "EXT" version should be removed.

scripts/core/subAllocationsProperties.yml
```diff
@@ -0,0 +1,62 @@
...
+base: $x_base_properties_t
+members:
+    - type: "const void*"
+      name: ptr
+      desc: "[out][optional] base address of the sub-allocation"
+    - type: size_t
+      name: size
+      desc: "[out][optional] size of the allocation"
+details:
+    - "This structure may be passed in $x_sub_allocations_properties_exp_desc_t, when querying for the properties of the sub-allocations of an allocation."
```
bashbaug, line 41: Can you check that this detail is saying what you'd like it to say?

* The spelling of `$x_sub_allocations_properties_exp_desc_t` doesn't look quite right - should this be `$x_sub_allocations_exp_desc_t` instead?
* Is this structure "passed in" or "returned from" the query?
* If the structure is purely returned from the query does it need to be derived from the base properties and hence have an stype and pNext or can this be omitted?
    * Note that there is an stype and a pNext for the sub-allocations descriptor `$x_sub_allocations_exp_desc_t` if we need to extend the chain to query other properties.

bashbaug (changes requested):
Although it is not required you may want to consider adding some sample code showing how this extension can be used.

jandres742:
@bashbaug 

> Since this is currently an experimental extension I think we only need the "EXP version below and this "EXT" version should be removed.

correct. done.

> The spelling of $x_sub_allocations_properties_exp_desc_t doesn't look quite right - should this be $x_sub_allocations_exp_desc_t instead?

correct. done.

> Is this structure "passed in" or "returned from" the query?

it is an array, as we have other queries that use pCount. I have added a sample usage so it is clearer.

bashbaug:
Thanks, the example is consistent with the spec now, but now I'm going to try one more time to convince you that the stype and pNext are extraneous in this case and they should be removed :smiley:.

We already have an stype and pNext for `ze_sub_allocations_exp_desc_t`, to identify and allow extending the structure passed into this query.  Do we really need another one for stype and pNext for `ze_sub_allocations_exp_properties_t`, which is the data returned from the query?

This is only an experimental extension so I'm OK merging this as-is, but I'd still prefer to get this as right as possible.

jandres742:
@bashbaug . Agree, in terms of actually needing it for exp_properties, we dont. They are base of base_properties, which is why we added them. The one actually needed is the one for exp_desc here.

so if what you are saying is that we should eliminate the definition of SUB_ALLOCATIONS_EXP_PROPERTIES. Or are you saying to we should not have x_sub_allocations_exp_properties_t being a child of x_base_properties_t?

bashbaug:
Both of these, actually:

> so if what you are saying is that we should eliminate the definition of SUB_ALLOCATIONS_EXP_PROPERTIES. Or are you saying to we should not have x_sub_allocations_exp_properties_t being a child of x_base_properties_t?

We should only need one stype and pNext on the structure we're linking in.  We shouldn't need a stype or pNext on any of the "output" structures.

Note that naming-wise the structure we're linking in is currently a "desc" and the output structures are currently "properties", but this seems a little weird too.  I'd consider renaming the structure we're linking in to "properties" to match the `ze_memory_allocation_properties_t` structure it is linked into, and then choosing somethign different for the output structures, maybe something like `ze_sub_allocation_t` (no "desc" or "properties").

jandres742:
@bashbaugh: you are right. I was trying to make it align with other structs, but your suggestion is good. Please see latest changes.

scripts/core/EXT_Exp_SubAllocationProperties.rst
```diff
@@ -0,0 +1,70 @@
...
+    ${x}_memory_sub_allocations_exp_properties_t subAllocationDesc {};
+    uitn32_t numberOfSuballocations = 0;
+    subAllocationDesc.stype = ${X}_STRUCTURE_TYPE_MEMORY_SUB_ALLOCATIONS_PROPERTIES;
+    subAllocationDesc.pCount = &numberOfSuballocations;
+
+    ze_memory_allocation_properties_t memAllocProperties {};
+    memAllocProperties.stype = ${X}_STRUCTURE_TYPE_MEMORY_ALLOCATION_PROPERTIES;
+    memAllocProperties.pNext = &subAllocationDesc;
+
+    // call ${x}MemGetAllocProperties to get number of sub-allocations
```
bashbaug, line 54: This one is producing a warning when building the specs now.  I think I'd just spell it out `zeMemGetAllocProperties` vs. trying to link it with `${x}`.

> EXT_Exp_SubAllocationProperties.rst(54) : error : $xMemGetAllocProperties parameter count mismatch - 1 actual vs. 4 expected

scripts/core/EXT_Exp_SubAllocationProperties.rst
```diff
@@ -0,0 +1,70 @@
...
+.. parsed-literal::
+
+    ${x}MemAllocDevice(context, &desc, size, alignment, device, &ptr);
+
+    ${x}_memory_sub_allocations_exp_properties_t subAllocationDesc {};
+    uitn32_t numberOfSuballocations = 0;
+    subAllocationDesc.stype = ${X}_STRUCTURE_TYPE_MEMORY_SUB_ALLOCATIONS_PROPERTIES;
+    subAllocationDesc.pCount = &numberOfSuballocations;
+
+    ze_memory_allocation_properties_t memAllocProperties {};
```
bashbaug, line 50: It'd be good to link this one:

```suggestion
    ${x}_memory_allocation_properties_t memAllocProperties {};
```

scripts/core/common.yml
```diff
@@ -458,6 +458,10 @@ etors:
       desc: $x_fabric_edge_exp_properties_t 
       version: "1.4"
       value: "0x0002000C"
+    - name: MEMORY_SUB_ALLOCATIONS_PROPERTIES
```
bashbaug, line 461: Needs an EXP here.

```suggestion
    - name: MEMORY_SUB_ALLOCATIONS_EXP_PROPERTIES
```

bashbaug (changes requested):
Thanks, latest version looks good, just a few cosmetic issues to fix now and this will be ready to go.

scripts/core/EXT_Exp_SubAllocationProperties.rst
jandres742: @bashbaug thanks, fixed.

scripts/core/EXT_Exp_SubAllocationProperties.rst
jandres742: @bashbaug : thanks. Reworded comment.

scripts/core/common.yml
jandres742: @bashbaug thanks. missed it. Fixed.

scripts/core/EXT_Exp_SubAllocationProperties.rst
```diff
@@ -0,0 +1,70 @@
...
+each of those sub-allocations. The following pseudo-code demonstrates a basic use-case of this extension:
+
+.. parsed-literal::
+
+    ${x}MemAllocDevice(context, &desc, size, alignment, device, &ptr);
+
+    ${x}_memory_sub_allocations_exp_properties_t subAllocationDesc {};
+    uitn32_t numberOfSuballocations = 0;
+    subAllocationDesc.stype = ${X}_STRUCTURE_TYPE_MEMORY_SUB_ALLOCATIONS_EXP_PROPERTIES;
+    subAllocationDesc.pCount = &numberOfSuballocations;
```
bashbaug, line 48: One final comment - this doesn't need to change now, but it's something to consider, especially if this ever becomes a non-experimental extension: We debatably have an unnecessary level of indirection here.  Because this entire structure is already an in-out structure we could simply return the "count" here, rather than returning via a "pointer-to-count".

bashbaug (approved):
LGTM, thanks!

scripts/core/EXT_Exp_SubAllocationProperties.rst
jandres742: thanks @bashbaug . Yes, you are right there. We are following here the two-pass query model we have in other interfaces, but that we could remove one level of indirection at least. Yes, let's keep in mind depending on how this evolves based on feedback.

Issue #29, opened 2022-10-19
zeCommandQueueExecuteCommandLists and other API's doesn't allow to return ZE_RESULT_ERROR_OUT_OF_HOST_MEMORY / ZE_RESULT_ERROR_OUT_OF_DEVICE_MEMORY 


MichalMrozek:
zeCommandQueueExecuteCommandLists and other API's doesn't allow to return ZE_RESULT_ERROR_OUT_OF_HOST_MEMORY / ZE_RESULT_ERROR_OUT_OF_DEVICE_MEMORY. There are use cases where such error are appropriate, so those calls should allow to return those errors.

jandres742:
thanks @MichalMrozek .  I think this is very needed. We could put it in 1.5

@bashbaug : Let me know what you think. I can prepare PR.

jandres742:
also @bashbaug : do we want this error code on others (or maybe all) calls?

bashbaug:
Yeah, I think we'll want to have `ZE_RESULT_ERROR_OUT_OF_HOST_MEMORY` as a possible result value from all calls, just like `CL_OUT_OF_HOST_MEMORY`.  We should be able to do this in the scripts rather than modifying every API.

If we're OK having `ZE_RESULT_ERROR_OUT_OF_DEVICE_MEMORY` on every API we could add it similarly.  If we want to be more selective it'll be a little more work.  For reference, OpenCL allows `CL_OUT_OF_RESOURCES` on many API calls but not all, though I'm not sure the selectivity buys much.

jandres742:
thanks @bashbaug . Marking as v1.5.

jandres742:
@bashbaug : Please see https://github.com/oneapi-src/level-zero-spec/pull/30

jandres742:
@bashbaug : I'm thinking we should add also ERROR_UNKONWN and ERROR_UNSUPPORTED as default errors for all APIs, to allow L0 driver to return an error that doesnt fit the description of any other, and when a given API has not yet been implemented.

bashbaug:
> I'm thinking we should add also ERROR_UNKNOWN and ERROR_UNSUPPORTED as default errors for all APIs, to allow L0 driver to return an error that doesnt fit the description of any other, and when a given API has not yet been implemented.

I don't think it makes sense to add these errors unconditionally:

* `ERROR_UNKNOWN` shouldn't ever happen in theory.  If an error condition is missing in the specification, we should add it.
[...]

PR #30, opened 2022-10-24
Add additional default errors

jandres742:
Add additional default errors

Add OUT_OF_HOST_MEMORY and OUT_OF_DEVICE_MEMORY errors as
default to all functions.

Resolves: #29

### DIFF
--- a/scripts/core/cmdlist.yml
+++ b/scripts/core/cmdlist.yml
@@ -73,9 +73,6 @@ params:
     - type: $x_command_list_handle_t*
       name: phCommandList
       desc: "[out] pointer to handle of command list object created"
-returns:
-    - $X_RESULT_ERROR_OUT_OF_HOST_MEMORY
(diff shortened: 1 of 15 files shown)


bashbaug (approved):
LGTM.

For completeness, note that this adds OUT_OF_HOST_MEMORY and OUT_OF_DEVICE_MEMORY as errors for all functions and not just those mentioned in the PR description ("all append functions and to zeCommandQueueExecuteCommandLists"), but I think this is the right thing to do.

jandres742:
thanks @bashbaug . what do you think about my comment here:
https://github.com/oneapi-src/level-zero-spec/issues/29#issuecomment-1295488187 

> I'm thinking we should add also ERROR_UNKONWN and ERROR_UNSUPPORTED as default errors for all APIs, to allow L0 driver to return an error that doesnt fit the description of any other, and when a given API has not yet been implemented.

PR #31, opened 2022-10-26
Update the lxml version requirement to 4.9.1

wdamon-intel:
Requirements failed to install due to stale lxml version; updating to 4.9.1.

### DIFF
--- a/third_party/requirements.txt
+++ b/third_party/requirements.txt
@@ -11,7 +11,7 @@ exhale==0.3.0
 idna==2.8
 imagesize==1.1.0
 Jinja2==2.10.3
-lxml==4.4.2
+lxml==4.9.1
 Mako==1.1.0
 MarkupSafe==1.1.1
 packaging==19.2


jandres742:
thanks @wdamon-intel ! question, why do we need 4.9.1? latest available in Ubuntu from official repo is 4.5.0
https://packages.ubuntu.com/search?keywords=python-lxml 

while on suse 15.4 is 4.7.1
https://software.opensuse.org/package/python-lxml

do we need something specific from 4.9.1 or is just that you used latest from https://lxml.de/installation.html?

wdamon-intel:
We do not need anything specific from 4.9.1; I just used the latest version listed on the lxml website. Note, this seemed to work correctly on Ubuntu 22.04 LTS.

jandres742:
thanks @wdamon-intel . Do you think it would be possible to post the PR with 4.5.0, just to require what is already available in the official repos?

jandres742:
thanks @wdamon-intel . Which error do you see, could you share?

4.9.1 is latest release from source, but not from available official repos, so I'm just trying to see how to minimize burden on users forcing them to install dependencies that are not coming in the official repos.

wdamon-intel:
@jandres742 Here is an example from Windows, though Linux issues a similar error for a different file (`libxml/xmlversion.h`).

```
C:\Users\<username>\AppData\Local\Temp\xmlXPathInitsrt1nyno.c(1): fatal error C1083: Cannot open include file: 'libxml/xpath.h': No such file or directory
      error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Professional\\VC\\Tools\\MSVC\\14.29.30133\\bin\\HostX86\\x64\\cl.exe' failed with exit code 2
      *********************************************************************************
      Could not find function xmlCheckVersion in library libxml2. Is libxml2 installed?
      *********************************************************************************
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: legacy-install-failure

× Encountered error while trying to install package.
╰─> lxml

note: This is an issue with the package mentioned above, not pip.
```

jandres742:
thanks @wdamon-intel . ok, let's merge it. I tested it on my side on Linux w/o any issues.

Issue #7, opened 2022-08-14
Define set of interfaces to programmatically use SYSMAN interfaces with Core APIs


jandres742:
There are two issues we would like to address here:

As mentioned in https://github.com/oneapi-src/level-zero/issues/36, there's a desired to have an API to enable SYSMAN, instead of an environment variable.

Additionally, there are frequent requests to add to Core API an interface that provides similar functionality to one already existing on SYSMAN. An example is `zesMemoryGetState`, which provides info on available memory in the system. Adding an interface on Core for something that is already provided on SYSMAN duplicates code, and is not scalable in the long run. So it is better to allow for SYSMAN interfaces to be used with Core when the application needs it.

To provide a solution to these two issues, a new set of interfaces is being proposed to enable and access SYSMAN functionality. Idea is to provide interfaces to enable SYSMAN, as well as to discover SYSMAN devices, which will be used to access underlying functionality.

Initial proposal is the following, where we have `zesInit`, `zesDriverSupported`, and `zesGetDevice`.

```cpp
function main( ... )
    if (zeInit(0) != ZE_RESULT_SUCCESS || zesInit() != ZE_RESULT_SUCCESS)
        output("Can't initialize the API")
    else
        # Discover all the drivers
        uint32_t driversCount = 0
        zeDriverGet(&driversCount, nullptr)
        ze_driver_handle_t* allDrivers = allocate(driversCount * sizeof(ze_driver_handle_t))
        zeDriverGet(&driversCount, allDrivers)

        ze_driver_handle_t hDriver = nullptr
        for(i = 0 .. driversCount-1)
            if (zesDriverSupported(allDrivers[i]) == false)
                continue

            # Discover devices in a driver
            uint32_t deviceCount = 0
            zeDeviceGet(allDrivers[i], &deviceCount, nullptr)

            ze_device_handle_t* allDevices =
                allocate_memory(deviceCount * sizeof(ze_device_handle_t))
            zeDeviceGet(allDrivers[i], &deviceCount, allDevices)

            for(devIndex = 0 .. deviceCount-1)
                ze_device_properties_t device_properties
                zeDeviceGetProperties(allDevices[devIndex], &device_properties)
                if(ZE_DEVICE_TYPE_GPU != device_properties.type)
                    next
                # Get the Sysman device handle. Returns null if device is not supported.
                zes_device_handle_t hSysmanDevice = zesGetDevice(allDevices[devIndex])
[...]

PR #34, opened 2022-11-02
separate APIs for initializing and enumerating sysman

bashbaug:
This PR has several changes related to #7:

* It adds a `zesInit` function to initialize sysman separate from the `zeInit` function.  It does not require an environment variable.
* It adds separate `zesDriverGet` and `zesDeviceGet` functions to enumerate sysman drivers and devices, rather than enumerating devices via core APIs and casting to sysman device handles.
* After enumerating device handles, sysman can be used exactly as it was before.

I added the APIs and I also updated the programming guide to describe how they are used.

Things we still need to decide (could be fixed in this PR or separately):

* Do we want query functions to determine if a core driver or device handle can be converted to a sysman driver or device handle, e.g. `zesDriverSupported`, or are the separate functions to enumerate sysman drivers and devices sufficient?
* Likewise, do we want a function to convert between a core driver or device handle to a sysman driver or device handle?

### DIFF
--- a/scripts/sysman/PROG.rst
+++ b/scripts/sysman/PROG.rst
@@ -24,7 +24,13 @@ High-level overview
 Environment Variables
 ---------------------
 
+%if ver >= 1.5:
+The System Resource Management library may now be initialized without using environment variables by calling ${s}Init.
+
+For compatibility, the following environment variables may also be enabled during initialization for the respective feature.
+%else:
 The following environment variables are required to be enabled during initialization for the respective feature.
+%endif
 
 ## --validate=off
 +-----------------+-------------------------------------+------------+-----------------------------------------------------------------------------------+
@@ -39,12 +45,51 @@ The following environment variables are required to be enabled during initializa
 Initialization
 --------------
 
+%if ver >= 1.5:
+An application wishing to manage power and performance for devices may
+use the System Resource Management library to enumerate system management
+driver and device handles.
+
+The following pseudo-code demonstrates a basic initialization and device discovery sequence:
+
+.. parsed-literal::
+
+   function main( ... )
+       if (${s}Init(0) != ${X}_RESULT_SUCCESS)
+           output("Can't initialize the API")
+       else
+           # Discover all the drivers
+           uint32_t driversCount = 0
+           ${s}DriverGet(&driversCount, nullptr)
+           ${s}_driver_handle_t* allDrivers = allocate(driversCount * sizeof(${s}_driver_handle_t))
+           ${s}DriverGet(&driversCount, allDrivers)
+
+           ${s}_driver_handle_t hDriver = nullptr
+           for(i = 0 .. driversCount-1)
+               # Discover devices in a driver
+               uint32_t deviceCount = 0
+               ${s}DeviceGet(allDrivers[i], &deviceCount, nullptr)
+
+               ${s}_device_handle_t* hSysmanHandles =
+                   allocate_memory(deviceCount * sizeof(${s}_device_handle_t))
+               ${s}DeviceGet(allDrivers[i], &deviceCount, hSysmanHandles)
+
+               # Use the hSymanHandles to manage the devices
+
+       free_memory(...)
+
+For compatibility, an application may also use the Level0 Core API to
+enumerate through available accelerator devices in the system. For
+each device handle, an application can cast it to a sysman device handle
+to manage the system resources of the device.
+%else:
 An application wishing to manage power and performance for devices first
 needs to use the Level0 Core API to enumerate through available
 accelerator devices in the system and select those of interest.
 
 For each selected device handle, applications can cast it to a
 **Sysman device handle** to manage system resources of the device.
+%endif
 
 .. image:: ../images/tools_sysman_object_hierarchy.png
 
--- a/scripts/sysman/device.yml
+++ b/scripts/sysman/device.yml
@@ -10,6 +10,33 @@ type: header
 desc: "Intel $OneApi Level-Zero Tool APIs for System Resource Management (Sysman) - Device management"
 ordinal: "2"
 --- #--------------------------------------------------------------------------
+type: function
+desc: "Retrieves sysman devices within a sysman driver"
+version: "1.5"
+class: $sDevice
+name: Get
+decl: static
+ordinal: "0"
+details:
+    - "Multiple calls to this function will return identical sysman device handles, in the same order."
+    - "The application may call this function from simultaneous threads."
+    - "The implementation of this function should be lock-free."
+params:
+    - type: $s_driver_handle_t
+      name: hDriver
+      desc: "[in] handle of the sysman driver instance"
+    - type: "uint32_t*"
+      name: pCount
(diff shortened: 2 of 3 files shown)


scripts/sysman/driver.yml
```diff
@@ -0,0 +1,70 @@
...
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Initialize $OneApi System Resource Management (sysman)"
+version: "1.5"
+class: $s
+name: Init
+decl: static
+ordinal: "0"
+details: 
+    - "The application must call this function or $xInit with the $S_ENABLE_SYSMAN environment variable set before calling any other sysman function."
```
jandres742, line 30: @bashbaug : I think we should clarify that zeInit is still needed or not. what do you think? something like:

"Applications wishing to use Core APIs alongside sysman functions should call zeInit as well".

I dont think we have restrictions on which should be called first, whether zeInit or zesInit.

scripts/sysman/driver.yml
```diff
@@ -0,0 +1,70 @@
...
+type: function
+desc: "Retrieves sysman driver instances"
+version: "1.5"
+class: $sDriver
+name: Get
+decl: static
+ordinal: "0"
+details:
+    - "A sysman driver represents a collection of physical devices."
+    - "Multiple calls to this function will return identical sysman driver handles, in the same order."
```
jandres742, line 55: @bashbaug : should we clarify behavior of the affinity mask with respect to this order?

scripts/sysman/driver.yml
```diff
@@ -0,0 +1,70 @@
...
+    - "The application may pass nullptr for pDrivers when only querying the number of sysman drivers."
+    - "The application may call this function from simultaneous threads."
+    - "The implementation of this function should be lock-free."
+params:
+    - type: "uint32_t*"
+      name: pCount
+      desc: |
+            [in,out] pointer to the number of sysman driver instances.
+            if count is zero, then the loader shall update the value with the total number of sysman drivers available.
```
jandres742, line 64: @bashbaug : "the loader" or "the driver"?

scripts/sysman/driver.yml
```diff
@@ -0,0 +1,70 @@
...
+            if count is greater than the number of sysman drivers available, then the loader shall update the value with the correct number of sysman drivers available.
```
jandres742, line 65: Same.

scripts/sysman/driver.yml
```diff
@@ -0,0 +1,70 @@
...
+    - type: "$s_driver_handle_t*"
+      name: phDrivers
+      desc: |
+            [in,out][optional][range(0, *pCount)] array of sysman driver instance handles.
+            if count is less than the number of sysman drivers available, then the loader shall only retrieve that number of sysman drivers.
```
jandres742, line 70: Same.

scripts/sysman/PROG.rst
```diff
@@ -39,12 +45,51 @@ The following environment variables are required to be enabled during initializa
...
+                   allocate_memory(deviceCount * sizeof(${s}_device_handle_t))
+               ${s}DeviceGet(allDrivers[i], &deviceCount, hSysmanHandles)
+
+               # Use the hSymanHandles to manage the devices
+
+       free_memory(...)
+
+For compatibility, an application may also use the Level0 Core API to
+enumerate through available accelerator devices in the system. For
+each device handle, an application can cast it to a sysman device handle
```
jandres742, line 83: @bashbaug : can the opposite be done? cast a sysman device to a core device? this goes to the same as below, whether zesInit is enough or we need to call zeInit as well.

scripts/sysman/driver.yml
jandres742: @jitendrasharma1989 : Agree. That's why we need to clarify that.

bashbaug: I think the specification is correct as-written:

* Applications must call zesInit (or zeInit with ZES_ENABLE_SYSMAN) before calling any sysman functions.  This requirement is documented here.
* Applications must call zeInit before calling any core API functions.  This requirement is documented as part of zeInit.
* There is no ordering requirement.

I don't mind adding another line cautioning that this function does NOT initialize the core APIs, just to remove possible confusion, but I do think it is extraneous.

Regarding why we still support initializing sysman with zeInit + ZES_ENABLE_SYSMAN, this is because removing this functionality would break backwards compatibility so we cannot do it in a minor API revision.  This is something we should do when we make a major API revision, though.

scripts/sysman/driver.yml
bashbaug: Yes, I think so.  IIRC we do NOT want the affinity mask to apply to sysman handles, correct?

scripts/sysman/driver.yml
bashbaug: "the loader" is correct, because the loader manages the drivers.  We have the same text for zeDriverGet.

scripts/sysman/driver.yml
jandres742: thanks @bashbaug . I would say then to just add the line about the ordering requirement. Wording about zeInit vs zesInit calls is good then as you say.

scripts/sysman/PROG.rst
bashbaug: No, the opposite should not be considered safe, even if may currently work in practice.

jandres742: thanks @bashbaug . So this aligns with what we said above about the affinity mask. the affinity mask would apply to Core, and therefore, the Core devices would be a subset of the sysman devices, so you could cast a core device to sysman one (you would always have one), but not the other way around, since you may have masked out the core device associated for the sysman device.

scripts/sysman/driver.yml
bashbaug: I added two additional lines for clarification:

> * This function will only initialize sysman. To initialize other functions, call zeInit.

> * There is no requirement to call this function before or after zeInit.

scripts/sysman/driver.yml
bashbaug: I added the line:

> * The number and order of handles returned from this function is NOT affected by the ZE_AFFINITY_MASK or ZE_ENABLE_PCI_ID_DEVICE_ORDER environment variables.

I can add back support for ZE_ENABLE_PCI_ID_DEVICE_ORDER if needed, but I'd prefer to have a clean separation if possible.

scripts/sysman/device.yml
```diff
@@ -10,6 +10,34 @@ type: header
...
 --- #--------------------------------------------------------------------------
+type: function
+desc: "Retrieves sysman devices within a sysman driver"
+version: "1.5"
+class: $sDevice
+name: Get
+decl: static
+ordinal: "0"
+details:
+    - "Multiple calls to this function will return identical sysman device handles, in the same order."
```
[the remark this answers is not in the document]
bashbaug: Do you have an order in mind?

If it turns out this is required I'm inclined to do the ordering in a layer.

PR #38, opened 2022-11-10
Move image view extension to standard

jandres742:
This is being used by several customers, and feedback is that it is good to be promoted to standard.

Extension will still be kept to facilitate transition.

### DIFF
--- a/scripts/core/common.yml
+++ b/scripts/core/common.yml
@@ -410,6 +410,9 @@ etors:
     - name: DEVICE_IP_VERSION_EXT
       desc: $x_device_ip_version_ext_t
       version: "1.5"
+    - name: IMAGE_VIEW_PLANAR_EXT_DESC
+      desc: $x_image_view_planar_ext_desc_t
+      version: "1.5"
     - name: RELAXED_ALLOCATION_LIMITS_EXP_DESC
       desc: $x_relaxed_allocation_limits_exp_desc_t
       value: "0x00020001"
--- a/scripts/core/EXT_Exp_ImageViewPlanar.rst
+++ b/scripts/core/EXT_Exp_ImageViewPlanar.rst
@@ -14,6 +14,10 @@ from templates import helper as th
  Image View Planar Extension
 =============================
 
+%if ver >= 1.5:
+This experimental extension is deprecated and replaced by the ${X}_extension_image_view_planar standard extension.
+%endif
+
 API
 ----
 
--- a/scripts/core/EXT_ImageView.rst
+++ b/scripts/core/EXT_ImageView.rst
@@ -8,15 +8,11 @@ from templates import helper as th
 %>
 :orphan:
 
-.. _ZE_experimental_image_view:
+.. _ZE_extension_image_view:
 
 =========================
  Image View Extension
 =========================
-
-%if ver >= 1.5:
-This experimental extension is deprecated and replaced by the ${X}_extension_image_view standard extension.
-%endif
 
 API
 ----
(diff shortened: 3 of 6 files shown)


bashbaug (commented):
Can you please check that all necessary changes are included in this PR?

Some things to double-check:

* The extension name needs to include "extension".
* The extension structions need to include "ext".
* We need to define structure type enums for the new structures.
* We either need to define a new EXT_ImageView.rst file or rename/adapt the existing EXT_Exp_ImageView.rst to describe the now-standard extension.

Is this PR just for the "image view" extension or is it also intending to include the "image view planar" extension as well?

jandres742:
@bashbaug :
> * The extension name needs to include "extension".

We are here moving the experimental extension to standard interface. Do you want to move it to standard extension instead?

> Is this PR just for the "image view" extension or is it also intending to include the "image view planar" extension as well?

it is for both.

bashbaug:
> We are here moving the experimental extension to standard interface. Do you want to move it to standard extension instead?

I don't have a strong opinion about a standard extension vs. a part of the core specification.  I think this mostly depends whether we want this functionality to be optionally supported (in which case it should probably remain an extension) or whether it should be required (in which case it should be part of the core specification).

If it's part of the core specficiation though, I think it should move out the extension YAML file(s) and into the core specification files.  Note that not all parts will need to move, for example there is no need to have the "name" or "version" in the core specfication.

jandres742:
thanks @bashbaug . Considering what you are saying about having to be required, then i think it is better to keep it as extension, since we should think more if it should be required (this started as an experimental extension, which means from the beginning when we added images APIs we didnt even think this was required).

scripts/core/common.yml
```diff
@@ -462,6 +462,10 @@ etors:
       desc: $x_memory_sub_allocations_exp_properties_t
       version: "1.5"
       value: "0x0002000D"
+    - name: IMAGE_VIEW_PLANAR_EXT_DESC
+      desc: $x_image_view_planar_ext_desc_t
+      value: "0x0002000E"
+      version: "1.5"
```
bashbaug, line 468: I don't think we want to mix standard and experimental extensions like this so I would recommend the following:

1. Move the definition of this enum up before any of the "experimental" structure types, so just after `DEVICE_IP_VERSION_EXT`.
2. Do not assign the explicit enum value `0x0002000E` and use the automatic enum value instead.

bashbaug (commented):
I have one specific comment and two general comments.  The general comments are:

1. I think we also need to define landing pages for the standard extensions `EXT_ImageView` and `EXT_ImageViewPlanar`.  These can more-or-less be copies of the experimental extension landing pages `EXT_Exp_ImageView` and `EXT_Exp_ImageViewPlanar`, for now at least.
2. Do we want to deprecate the experimental extensions and/or signpost to the standard extension instead?  Presumably we'll want to remove the experimental extensions at some point, after applications have had a chance to switch over to the standard extensions.

jandres742:
> I have one specific comment and two general comments. The general comments are:
> 
> 1. I think we also need to define landing pages for the standard extensions `EXT_ImageView` and `EXT_ImageViewPlanar`.  These can more-or-less be copies of the experimental extension landing pages `EXT_Exp_ImageView` and `EXT_Exp_ImageViewPlanar`, for now at least.
> 2. Do we want to deprecate the experimental extensions and/or signpost to the standard extension instead?  Presumably we'll want to remove the experimental extensions at some point, after applications have had a chance to switch over to the standard extensions.

@bashbaug : Good idea. Don for both.

bashbaug (commented):
Getting closer!  Really, only the extraneous character is a must-fix.

bashbaug (approved):

PR #44, opened 2022-11-17
Add version added into @details if available

bmyates:

### DIFF
--- a/scripts/templates/helper.py
+++ b/scripts/templates/helper.py
@@ -768,6 +768,8 @@ def make_details_lines(namespace, tags, obj):
                 for line in split_line(subt(namespace, tags, item, True), 70):
                         lines.append("%s %s"%(prologue, line))
                         prologue = "     "
+    if 'version' in obj:
+        lines.append("    - Added in specification v%s" %obj['version'])
     if 'analogue' in obj:
         lines.append("")
         lines.append("@remarks")


bmyates:
idea is to put the version added into documentation when available. 

This still needs some format improvement and to make sure all cases are handled okay.

scripts/templates/helper.py
```diff
@@ -768,6 +768,8 @@ def make_details_lines(namespace, tags, obj):
                 for line in split_line(subt(namespace, tags, item, True), 70):
                         lines.append("%s %s"%(prologue, line))
                         prologue = "     "
+    if 'version' in obj:
+        lines.append("    - Added in specification v%s" %obj['version'])
```
jandres742, line 772: Thanks @bmyates . Do we actually need this? this will create a lot of verbosity in the documentation, no?

FYI: @bashbaug .

bmyates: It's adding one line per API for APIs that we have accurate version information for.   Not all APIs have the version included in the yaml.   For example - none of the original APIs from 1.0 have version information.   Some of the APIs added after 1.0 don't have version information either.

If we agree it is good to add this information, then we should fine-tune how/where we want to present this in the html.

bmyates:
closing due to no activity

PR #48, opened 2022-12-02
fix scripts to enable building older spec versions

bashbaug:
Fixes several issues affecting older spec builds:

* The scripts do not handle "else" statements.  Fixed by inverting the "if" condition instead.
* The scripts were not properly handling "endif" statements.  Fixed by correcting the script regular expression.

### DIFF
--- a/scripts/core/PROG.rst
+++ b/scripts/core/PROG.rst
@@ -1096,7 +1096,7 @@ A kernel timestamp event is a special type of event that records device timestam
 %if ver >= 1.1:
        const double timestampFreq = NS_IN_SEC / device_properties.timerResolution;
 %endif
-%if ver < 1.5:
+%if ver < 1.1:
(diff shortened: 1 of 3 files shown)


scripts/core/PROG.rst
```diff
@@ -1094,7 +1095,8 @@ A kernel timestamp event is a special type of event that records device timestam
        // Get timestamp frequency
 %if ver >= 1.1:
        const double timestampFreq = NS_IN_SEC / device_properties.timerResolution;
-%else:
+%endif
+%if ver < 1.5:
```
jandres742, line 1099: @bashbaug : should it be less than 1.1?

bashbaug: Eek, yes, too much copy-paste.  Fixed!

jandres742 (approved):

Issue #51, opened 2022-12-12
clarification on zetCommandListAppendMetricQueryEnd phWaitEvents


jandres742:
For zetCommandListAppendMetricQueryEnd   https://spec.oneapi.io/level-zero/latest/tools/api.html?highlight=zetcommandlistappendmetricqueryend#zetcommandlistappendmetricqueryend we have 

phWaitEvents – [in][mbz] must be nullptr

[ZE_RESULT_ERROR_INVALID_NULL_POINTER](https://spec.oneapi.io/level-zero/latest/core/api.html#ze__api_8h_1a3ea791bcf22ddfc1d413579b12c22767a90dc7f955a711ba2377da3125672400f)

nullptr == phWaitEvents

so it seems we have contradictory statements. phWaitEvents must be zero, but if so, we return INVALID_NULL_POINTER. We need to choose one of the two.

jandres742:
@bmyates : do you remember why it was changed intentionally to mbz in 1.1? if there was a reason at that moment, do we know if that requirement does not longer apply?

bmyates:
Requirements are unchanged.  Original customer report that motivated change: 

>  zetCommandListAppendMetricQueryBegin has no hSignalEvent or phWaitEvents parameters whereas zetCommandListAppendMetricQueryEnd does. If zetCommandListAppendMetricQueryBegin is a barrier, the specification doesn't say so.
> MetricQueryEnd should only have a signal and not wait. Doesn't hurt that wait is here but not useful. MetricQueryBegin doesn’t need a signal or wait. Need to be defined as barriers.

jandres742:
changes merged in https://github.com/oneapi-src/level-zero-spec/pull/52

PR #52, opened 2022-12-12
remove nullptr error code from params with mbz trait

bmyates:


jandres742:
Thanks @bmyates . Let's clarify why this is in the first place here. Discussion in #51.

jandres742 (changes requested):
Pending discussion in #51

jandres742:
clarification in https://github.com/oneapi-src/level-zero-spec/issues/51 looks good to me.
@bashbaug , @zackwaters : let me know if we want to squeeze this on 1.5. I dont see a risk, since mbz is only used by metricQueryEnd.

jandres742 (approved):

PR #62, opened 2023-01-16
Fix ze_device_properties_t in samples

jandres742:
Correct initialize ze_device_properties_t to zero and set stype properly. This way samples show how to initialize these structs to avoid driver implementations to misread garbage values in pNext.


jandres742:
FYI: @wdamon-intel

wdamon-intel (approved):

PR #63, opened 2023-01-17
Extend the SYSMAN Frequency Domain list to include a MEDIA Domain

pmmeyer:
Addresses #60.

### DIFF
--- a/scripts/sysman/frequency.yml
+++ b/scripts/sysman/frequency.yml
@@ -21,6 +21,7 @@ etors:
       desc: "Local Memory Domain."
     - name: MEDIA
       desc: "GPU Media Domain."
+      version: "1.6"
 --- #--------------------------------------------------------------------------
 type: struct
 desc: "Frequency properties"
--- a/scripts/sysman/frequency.yml
+++ b/scripts/sysman/frequency.yml
@@ -19,6 +19,8 @@ etors:
       desc: "GPU Core Domain."
     - name: MEMORY
       desc: "Local Memory Domain."
+    - name: MEDIA
+      desc: "GPU Media Domain."
 --- #--------------------------------------------------------------------------
 type: struct
 desc: "Frequency properties"


jandres742:
@pmmeyer : thanks. This addresses https://github.com/oneapi-src/level-zero-spec/issues/60, correct?

@sanilkumar0 , @saik-intel: could you take a look?

pmmeyer:
> @pmmeyer : thanks. This addresses #60, correct?
> 
> @sanilkumar0 , @saik-intel: could you take a look?

Yes, that is correct.

ezhilintc:
> > @pmmeyer : thanks. This addresses #60, correct?
> > @sanilkumar0 , @saik-intel: could you take a look?
> 
> Yes, that is correct.

This addresses spec side of it. We still need some changes in sysman UMD, right?

jandres742:
@ezhilintc : Correct. Once these changes are merged and released, we could start implementation in driver.

I think this could be added in next minor version, v1.6.

ezhilintc (approved):

pmmeyer:
@jandres742 - will someone be merging this PR? I don't seem to have permission.

jandres742:
@pmmeyer thanks. We haven't merged this PR since we are targeting this for v1.6, since it is a change in enum. Before that, we are thinking on having another update, 1.5.X, so we will merge this afterwards. After merging, it will be released on 1.6 loader, and then absorbed by driver implementations.

Let me know if this helps.

jandres742 (changes requested):
To be merged for v1.6.

scripts/sysman/frequency.yml
```diff
@@ -19,6 +19,8 @@ etors:
       desc: "GPU Core Domain."
     - name: MEMORY
       desc: "Local Memory Domain."
+    - name: MEDIA
```
bmyates, line 22: Add a `version: 1.6` guard to new addition

bmyates:
Code targeting 1.6 should be mergeable if it is properly guarded with version checks

jandres742:
> Code targeting 1.6 should be mergeable if it is properly guarded with version checks

thanks @bmyates . You are correct. @pmmeyer : could you add the versioning?

wdamon-intel:
> could you add the versioning

@pmmeyer please follow up with the versioning tags.

jandres742 (approved):

PR #66, opened 2023-02-06
Explicitly state the timestamp unit for the memory bandwidth API

### LINKED ISSUE #16: Timestamp unit is not specified for all Sysman metric types
Timestamp unit / resolution is explicitly specified as microseconds for following metrics:
* https://spec.oneapi.io/level-zero/latest/sysman/api.html#zes-pci-stats-t
* https://spec.oneapi.io/level-zero/latest/sysman/api.html#zes-engine-stats-t
* https://spec.oneapi.io/level-zero/latest/sysman/api.html#zes-fabric-port-throughput-t
* https://spec.oneapi.io/level-zero/latest/sysman/api.html#zes-freq-throttle-time-t
* https://spec.oneapi.io/level-zero/latest/sysman/api.html#zes-power-energy-counter-t

But not for:
* https://spec.oneapi.io/level-zero/latest/sysman/api.html#zes-mem-bandwidth-t

Same microsecond resolution is implied by the provided bandwidth algorithm, but it should be _explicitly_ stated also for the memory bandwidth timestamp.

----

PS. Maybe there could be also some generic statement at start of the docs, about all timestamps in the Sysman API being reported as microseconds, regardless of what resolution HW actually offers for the backend.

Or is it possible that the nanoseconds and cycles mentioned in core API (https://spec.oneapi.io/level-zero/latest/core/api.html#ze-device-properties-t) could be used in Sysman API later on?

wdamon-intel:
Resolves #16 

### DIFF
--- a/scripts/sysman/memory.yml
+++ b/scripts/sysman/memory.yml
@@ -150,7 +150,7 @@ members:
     - type: uint64_t
       name: timestamp
       desc: |
-            [out] The timestamp when these measurements were sampled.
+            [out] The timestamp in microseconds when these measurements were sampled.
             This timestamp should only be used to calculate delta time between snapshots of this structure.
             Never take the delta of this timestamp with the timestamp from a different structure since they are not guaranteed to have the same base.
             The absolute value of the timestamp is only valid during within the application and may be different on the next execution.


scripts/sysman/memory.yml
```diff
@@ -150,7 +150,7 @@ members:
     - type: uint64_t
       name: timestamp
       desc: |
-            [out] The timestamp when these measurements were sampled.
+            [out] The timestamp in microseconds when these measurements were sampled.
```
MichalMrozek, line 153: how it is calculated?
In other places (i.e. for event profiling ) we have timestamp resolution in nanoseconds, microseconds may be too coarse.

eero-t: How often you're polling it?

It is in microseconds for every other metric, see: https://github.com/oneapi-src/level-zero-spec/issues/16

wdamon-intel: Is it expected that the timestamp resolution for any/all metrics be defined by the `timerResolution` device property? If so, we should change the spec language more generally.

eero-t: > See here -> https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=timerresolution#_CPPv4N22ze_device_properties_t15timerResolutionE
> ...
> Another example is synchronize call
https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=timerresolution#zeeventhostsynchronize

But those are L0 core APIs?   Memory bandwidth is L0 Sysman API.

All _Sysman_ API metric timestamps (other than the currently undefined mem BW) are already documented to be in microseconds, and e.g. Intel's public Sysman applications, such as XPU manager and collectd Sysman plugin already handle them as such (have done so since compute runtime backend originally started to support those metrics).

PS. If those Sysman metric APIs are going to change the reported timestamp resolution, that is an API change which needs to go through appropriate API change process (publicly announce deprecation of the old APIs, provide new ones, and eventually drop the deprecated ones after suitable period of supporting both).

And there better be some mechanism through which frontend, backend and applications can detect each others' versions, and act accordingly, depending on which L0 frontend version app happens to be built against, and with which backend version that happens to run with.

jandres742: @MichalMrozek : to clarify here: this PR is changing the metrics in SYSMAN, not on Core.

MichalMrozek: Thanks for claryfing, what is the reason of having different time precision domains between SYSMAN and Core?

eero-t: It's a question of use-cases, their query intervals, and whether the values are counters (like errors), or sampled "gauge" values (like frequency).   If you poll often, you need more fine grained timestamp, and may care less about timestamps wrapping around between them.

AFAIK Sysman metrics are mainly for telemetry.  In a cluster that can have from tens to even thousands of nodes, polling of them is done at several seconds interval, maybe sometimes even at minute intervals (counters are obviously more useful for this than things that can only be sampled).  Otherwise telemetry system drowns in data, as GPU metrics are only small part of it.

Whereas for profiling e.g. GPU usage of individual functions, much shorter time intervals are needed to get meaningful results.  And that profiling is often focused on some specific types of metrics, so amount of data is less of a problem.

jandres742: > Thanks for claryfing, what is the reason of having different time precision domains between SYSMAN and Core?

@MichalMrozek : as @eero-t  mentions above, Metrics reported by SYSMAN and CORE may rely on different underlying HW and kernel support. For instance, for SYSMAN, it is more likely that resolution is determined by hardware counters or what the firmware has defined.

wdamon-intel: @sanilkumar0 confirmed timestamps are the same across SYSMAN API (CPU-based) and the resolution is in microseconds.

PR #67, opened 2023-02-06
Clarify language for pString parameter of zeKernelGetSourceAttributes

### LINKED ISSUE #32: Clarification on pString in zeKernelGetSourceAttributes
For zeKernelGetSourceAttributes we have pString as forbidden to be passed as null:

ZE_APIEXPORT [ze_result_t](https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=zekernelgetsourceattributes#_CPPv411ze_result_t) ZE_APICALL zeKernelGetSourceAttributes(ze_kernel_handle_t hKernel, uint32_t *pSize, char **pString)

[ZE_RESULT_ERROR_INVALID_NULL_POINTER](https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=zekernelgetsourceattributes#ze__api_8h_1a3ea791bcf22ddfc1d413579b12c22767a90dc7f955a711ba2377da3125672400f)

nullptr == pSize

nullptr == pString

However, what we usually do in this kind of interfaces is that the user first requests for pSize, then allocates the array's memory (in this case pString), and then in a second call passes that array to the driver, so the driver can copy the requested data.

@bashbaug : is my understanding correct, or is this a special case where the lifetime of the array is managed by the driver. That seems to be implied by this statement:

pString – [in,out] pointer to null-terminated string, whose lifetime is tied to the kernel object, where kernel source attributes are separated by space.

if so, shouldn't be pString just a pointer, instead of a double pointer?

wdamon-intel:
Resolves #32 

Signed-off-by: Will Damon <will.damon@intel.com>

### DIFF
--- a/scripts/core/module.yml
+++ b/scripts/core/module.yml
@@ -557,10 +557,13 @@ params:
       desc: "[in] handle of the kernel object"
     - type: "uint32_t*"
       name: pSize
-      desc: "[in,out] pointer to size of string in bytes."
+      desc: "[in,out] pointer to size of string in bytes, including null-terminating character."
     - type: "char**"
       name: pString
-      desc: "[in,out] pointer to null-terminated string, whose lifetime is tied to the kernel object, where kernel source attributes are separated by space."
+      desc: |
+            [in,out][optional] pointer to application-managed character array (string data).
+            If NULL, the string length of the kernel source attributes, including a null-terminating character, is returned in pSize.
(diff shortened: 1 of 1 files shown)


wdamon-intel:
@jandres742 please review

jandres742 (approved):
thanks @wdamon-intel ! looks good to me.

jandres742 (approved):

PR #69, opened 2023-02-07
Clarify intended interpretation of 32-bit device id.

### LINKED ISSUE #54: Clarify ze_device_properties_t::deviceId in the specs
Question from customer:
It is unclear from the wording in the spec how to treat the data. PCI ID is a 16-bit value, but L0 provides a 32-bit value: which exact 16 bits within a 32-bit value should be interpreted as PCI ID by a user

Reply:
The least-significant 16-bits of the 32-bit contain the device id

We need to clarify this in the spec.

wdamon-intel:
Resolves #54

### DIFF
--- a/scripts/core/device.yml
+++ b/scripts/core/device.yml
@@ -128,7 +128,9 @@ members:
       desc: "[out] vendor id from PCI configuration"
     - type: uint32_t
       name: deviceId
-      desc: "[out] device id from PCI configuration"
+      desc: |
+            [out] device id from PCI configuration
+            Note, the least-significant 16-bits contain the device id.
     - type: $x_device_property_flags_t
       name: flags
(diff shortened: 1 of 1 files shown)


wdamon-intel:
@jandres742 please review

scripts/core/device.yml
```diff
@@ -128,7 +128,9 @@ members:
       desc: "[out] vendor id from PCI configuration"
     - type: uint32_t
       name: deviceId
-      desc: "[out] device id from PCI configuration"
+      desc: |
+            [out] device id from PCI configuration
+            Note, the least-significant 16-bits contain the device id.
```
jandres742, line 133: @wdamon-intel : I think we could be more generic and say this uses little-endian format.

jandres742 (approved):

PR #71, opened 2023-02-07
Clarify that zeContextMakeMemoryResident is a cross-platform API

### LINKED ISSUE #70: Core programming guide incorrectly calls out zeContextMakeMemoryResident as Windows-only
`zeContextMakeMemoryResident` is not a Windows-only API, it is supported on Linux as well. The 'Windows-only' comment needs to be removed from the core programming guide.

wdamon-intel:
- Remove incorrect stipulation that `zeContextMakeMemoryResident` is Windows-only. It is a cross-platform API.
- Also, correct a spelling mistake.

Resolves #70

### DIFF
--- a/scripts/core/PROG.rst
+++ b/scripts/core/PROG.rst
@@ -1805,7 +1805,12 @@ such as multiple levels of indirection, there are two methods available:
 
     + If the driver is unable to make all allocations resident, then the call to ${x}CommandQueueExecuteCommandLists will return ${X}_RESULT_ERROR_OUT_OF_DEVICE_MEMORY
 
+%if ver >= 1.6:
 2. Explicit ${x}ContextMakeMemoryResident APIs are included for the application to dynamically change residency as needed.
+%endif
+%if ver < 1.6:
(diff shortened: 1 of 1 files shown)


scripts/core/PROG.rst
```diff
@@ -1805,7 +1805,7 @@ such as multiple levels of indirection, there are two methods available:
 
     + If the driver is unable to make all allocations resident, then the call to ${x}CommandQueueExecuteCommandLists will return ${X}_RESULT_ERROR_OUT_OF_DEVICE_MEMORY
 
-2. Explcit ${x}ContextMakeMemoryResident APIs are included for the application to dynamically change residency as needed. (Windows-only)
+2. Explicit ${x}ContextMakeMemoryResident APIs are included for the application to dynamically change residency as needed.
```
jandres742, line 1808: thanks @wdamon-intel . I would add the guards for 1.6 here. Previous to this change, an L0 driver on Linux wouldn't need to implement this API to be compliant, but now we are saying this needs to be implemented on Linux also to be compliant. So to not disrupt drivers that may have been already compliant before, better to have this requirement starting in the next minor version.

What do you think?

jandres742 (approved):

PR #77, opened 2023-02-21
Introduce extension to query normalized kernel event timestamps

### LINKED ISSUE #55: Define new interface for parsing the raw timestamps into time units.
L0 currently only has interfaces that return timestamps on its raw form. While this is useful for some tools, it would be of benefit if L0 also returns the timestamps mapped to the host, similar to clGetDeviceAndHostTimer. 
Idea here is to define a new interface for this.

wdamon-intel:
This extension enables querying of event timestamps with up to three views of the data: CPU normalized, GPU normalized and GPU raw. It is intended to compliment and eventually replace all usages of zeEventQueryTimestampsExp() and zeEventQueryKernelTimestamp().

Resolves #55 

Signed-off-by: Will Damon <will.damon@intel.com>


jandres742:
@aravindksg: this addresses https://github.com/oneapi-src/level-zero-spec/issues/55. could you check if it fulfills requirements?

scripts/core/eventQuerySynchronizedTimestamps.yml
```diff
@@ -0,0 +1,66 @@
+#
+# Copyright (C) 2021-2022 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+#
+# See YaML.md for syntax definition
+#
+--- #--------------------------------------------------------------------------
+type: header
+desc: "Intel $OneApi Level-Zero Extension for supporting event query synchronized timestamps."
```
aravindksg: minor. reword to "extension for supporting querying of synchronized event timestamps" ?

scripts/core/eventQuerySynchronizedTimestamps.yml
```diff
@@ -0,0 +1,66 @@
...
+      name: hDevice
+      desc: "[in] handle of the device to query"
+    - type: "uint32_t*"
+      name: pCount
+      desc: |
+            [in,out] pointer to the number of timestamp results.
+            if count is zero, then the driver shall update the value with the total number of timestamps available.
+            if count is greater than the number of timestamps available, then the driver shall update the value with the correct number of timestamps available.
+    - type: "uint64_t*"
+      name: pHostTimestamps
```
aravindksg: why would we want to return N number of host timestamps?

we also already have `zeDeviceGetGlobalTimestamps` which returns synchronized host and device timestmaps

scripts/core/EXT_EventQueryKernelTimestamps.rst
```diff
@@ -0,0 +1,79 @@
...
+
+    * ${x}_event_query_kernel_timestamps_ext_version_t
+
+
+Event Query Kernel Timestamps
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This extension enables the query of event timestamps to provide up to three views of the data, for a given event on a given device:
+
+1. Host timestamps, normalized with Device timestamps
```
aravindksg, line 37: why return host timestamps normalized to device timestamps? I think the user feedback was that they would like to query for event timestamps that return a) data in time domain (rather than raw timestamps as we have now) and b) to return timestamps normalized to host time scale (as OCL does it with `clGetEventProfilingData`). So points 2 and 3 below seem to address the feedback. Also as noted in earlier commit, we already have `zeDeviceGetGlobalTimestamps` which would address 1)

scripts/core/EXT_EventQueryKernelTimestamps.rst
```diff
@@ -0,0 +1,79 @@
...
+2. Device timestamps, normalized with Host timestamps
+3. Raw device timestamps
+
+Each view requires unique backing, provided by the caller.
+
+**Notes**
+
+- There is a slight performance implication to consider when requesting normalized timestamp data as the runtime must perform some calculations on behalf of the caller.
+- This extension is designed to compliment and eventually replace all usages of ${x}EventQueryTimestampsExp and ${x}EventQueryKernelTimestamp.
+
```
aravindksg, line 47: do we need to add notes on behavior for when user uses device handles vs sub-device handles?

scripts/core/eventQuerySynchronizedTimestamps.yml
wdamon-intel: Thanks for catching that. I meant to rephrase it and remove the word 'synchronized' altogether.

scripts/core/eventQuerySynchronizedTimestamps.yml
wdamon-intel: There may be more than one timestamp associated with an event. The `zeDeviceGetGlobalTimestamps` API is just that, a global timestamp read from the host, and is not associated with an event.

scripts/core/EXT_EventQueryKernelTimestamps.rst
wdamon-intel: I think perhaps the way this was phrased is not quite as accurate or precise as it should be. The intention is a) and b) above. Only the final parameter is intended to provide the caller with raw (device) timestamps, associated with the given event.

scripts/core/EXT_EventQueryKernelTimestamps.rst
wdamon-intel: Yes, I can add language for those conditions.

jandres742: > do we need to add notes on behavior for when user uses device handles vs sub-device handles?

@aravindksg : why kind of behavior are you thinking we need to explain?

scripts/core/EXT_EventQueryKernelTimestamps.rst
jandres742: @aravindksg so you suggest we just returned deviceTimestamps, which are mapped to host. Host timestamps mapped to device are not needed? (would they be?)

aravindksg: right, I'm not sure there's a customer need for host timestamps mapped to device time domain since the device operations are anyway host initiated. For host timings they could just as easily capture it themselves in upper levels. In this API, we mainly just need to provide device related info (but massage it per their feedback)

I wonder if there's a way we can define this new API with a `desc` which has a `pNext` such that if required we could extend the functionality of this API further based on customer need.

jandres742: if we extend this, would be with a properties, so instead of returning values in arguments, we could pass a props struct, and populate it, and use pNext for future extension of functionality.

@wdamon-intel : what do you think?

scripts/core/eventQuerySynchronizedTimestamps.yml
aravindksg: but events are attached to device operations (kernels/copy/fill), so we mainly just need to return device side timing info right?

jandres742: Same discussion as below. So yes, no need for hostTimestamps in this new API, right?

scripts/core/EXT_EventQueryKernelTimestamps.rst
aravindksg: off the cuff statement: "The driver could return different values for pCount when queried with device vs sub-device handles based on underlying optimizations performed. Therefore user's should take care to both query pCount and the timestamps with appropriate device/sub-device handles. Driver shall only return timestamp data that is available for a given handle if pCount is set incorrectly by user"

something along those lines

jandres742: pCount is up to the number of sub-devices, right? maybe indicating that is enough, like "pcount may be up to the number of sub-devices"?

scripts/core/eventQuerySynchronizedTimestamps.yml
aravindksg: yes

scripts/core/EXT_EventQueryKernelTimestamps.rst
aravindksg: `pCount` could be up to number of event packets that implementation needs to return. it could depend on different optimizations driver may do internally. so not necessarily `pCount == numSubdevices`

jandres742: thanks @aravindksg . You are right. we can add then:

"The value returned by pCount is implementation specific."

scripts/core/EXT_EventQueryKernelTimestamps.rst
```diff
@@ -0,0 +1,79 @@
...
+    ${x}EventQueryKernelTimestampsExt(hEvent, hDevice, &count, nullptr, nullptr, nullptr);
+
+    // Allocate storage for timestamp data
+    hostTimestamps.resize(count);
+    deviceTimestamps.resize(count);
+    rawDeviceTimestamps.resize(count);
+
+    // Query the event timestamps
+    ${x}EventQueryKernelTimestampsExt(hEvent, hDevice, &count,
+        hostTimestamps.data(),          /\* [optional] normalized event host timestamp data that correlates to device timestamps \*/
```
MichalMrozek, line 77: I would like to first have a dicsussion on how we are expected to implement this in Level Zero.
Right now we cannot easily transfer device timestamps to host timestamp, there is no easy mapping.
In OCL to do this, we need to query submit host timestamps and then compute delta's.
Querying host timestamp is massive cost as it requires going to KMD.
With current append call taking ~2us, this KMD trip is 7us, so call is 4x slower, this wouldn't be acceptable for our clients profiling their workloads. (too much overhead drastically changing the outcome of workload)

With this extension we would need to go to KMD for normal timestamp events , so we could service device -> host transition later, this would introduce humongous overhead that is not present today.

jandres742: thanks @MichalMrozek . Performance hit is mentioned above "There is a slight performance implication to consider when requesting normalized timestamp data as the runtime must perform some calculations on behalf of the caller."

Main objective here is to have parity with OCL, where users dont need to do all calculations to try to map to the host, so we would have parity on both API and implementation (if L0 would have overhead, it should be the same as OCL).

Because of the performance considerations, this API (and the other ones we already have) still return raw timestamps for applications that don't want to have that overhead and prefer to access the raw timestamps.

what do you think?

jandres742: @MichalMrozek could you elaborate on why would we tax all timestamps events? As I see, as long as users dont call this API, then all timestamps obtained with already-existing APIs would have same overhead, correct?

MichalMrozek: how would you then service this API and provide host timestamps for device timestamps?

jandres742: @MichalMrozek : could you elaborate on the question? I still think that any additional overhead that you mention, like extra calls to KMD, would be specific to this call, or isn't that the case?

MichalMrozek: If you call to KMD on this call, you receive current host & device pair of timestamps that is after the execution happened.

You cannot correlate those to device timestamps that you have recorded when the kernel was executed as you do not know how many times device timestamp (which is 32 bit) overflowed (it does it every 343 seconds), so you need to have host time recorded close to the time when event is signaled.

jandres742: thanks @MichalMrozek . is that the same thing OpenCL GPU driver implementation is doing?

is your suggestion then to have a new flag, something like ZE_EVENT_POOL_FLAG_KERNEL_MAPPED_TIMESTAMP, so we know ahead of time whether overhead is needed?

MichalMrozek: correct, this way applications that want to pay the full price can do this without affecting applications that doesn't want to pay for this.

wdamon-intel:
@jandres742 please have a look at the revised proposal reflected by the latest commits.

scripts/core/event.yml
```diff
@@ -448,7 +448,7 @@ class: $xCommandList
 name: AppendQueryKernelTimestamps
 details:
     - "The application must ensure the events are accessible by the device on which the command list was created."
-    - "The application must ensure the events were created from an event pool that was created using $X_EVENT_POOL_FLAG_KERNEL_TIMESTAMP flag."
+    - "The application must ensure the events were created from an event pool that was created using $X_EVENT_POOL_FLAG_KERNEL_TIMESTAMP or $X_EVENT_POOL_FLAG_KERNEL_MAPPED_TIMESTAMP flag."
```
aravindksg, line 451: do we need to add note here that synchronized timestamps would not be available via this `appendQueryKernelTimestmps` ?

scripts/core/eventQueryKernelTimestamps.yml
```diff
@@ -0,0 +1,103 @@
...
+      desc: |
+            [in,out] pointer to the number of event packets available.
+                - This value is implementation specific.
+                - if `*pCount` is zero, then the driver shall update the value with the total number of event packets available.
+                - if `*pCount` is greater than the number of event packets available, the driver shall update the value with the correct value.
+                - Buffer(s) for query results must be sized by the application to accommodate a minimum of `*pCount` elements.
+    - type: $x_event_query_kernel_timestamps_results_ext_properties_t*
+      name: pResults
+      desc: |
+            [in][optional] pointer to event query properties structure(s).
```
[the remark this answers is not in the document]
aravindksg: @joshuaranjan why is this out? it's a props struct

scripts/core/EXT_EventQueryKernelTimestamps.rst
```diff
@@ -64,8 +64,8 @@ This extension enables the querying of synchronized event timestamps.
     // Determine the level of support by getting the module properties
     ${x}DeviceGetProperties(hDevice, &devProps);
 
-    const bool supportsRaw = (0 != (tsProps.flags & ${X}_EVENT_QUERY_KERNEL_TIMESTAMPS_EXT_FLAG_RAW));
-    const bool supportsSynchronized = (0 != (tsProps.flags & ${X}_EVENT_QUERY_KERNEL_TIMESTAMPS_EXT_FLAG_HOST_SYNCHRONIZED));
+    const bool supportsKernel = (0 != (tsProps.flags & ${X}_EVENT_QUERY_KERNEL_TIMESTAMPS_EXT_FLAG_KERNEL));
+    const bool supportsSynchronized = (0 != (tsProps.flags & ${X}_EVENT_QUERY_KERNEL_TIMESTAMPS_EXT_FLAG_SYNCHRONIZED));
```
aravindksg, line 68: minor : reword variables for better clarity? example "supportsKernel" -> "supportsKernelTs"

scripts/core/EXT_EventQueryKernelTimestamps.rst
```diff
@@ -74,27 +74,27 @@ This extension enables the querying of synchronized event timestamps.
...
         // Get the number of timestamps associated with the event.
         ${x}EventQueryKernelTimestampsExt(hEvent, hDevice, &count, nullptr);
 
-        // Allocate storage for raw device timestamps
-        std::vector<${x}_kernel_timestamp_result_t> rawTimestamps(count);
+        // Allocate storage for kernel timestamp results
+        std::vector<${x}_kernel_timestamp_result_t> timestamps(count);
 
-        // Allocate storage for synchronized event timestamps, in nanoseconds
+        // Allocate storage for synchronized timestamp results, in nanoseconds
```
aravindksg, line 87: type for synchronized timestamps should also be `kernel_timestamps_result_t`?

scripts/core/event.yml
wdamon-intel: @aravindksg Would it be better to disallow events from a `KERNEL_MAPPED` pool?

aravindksg: @wdamon-intel yeah that might be a good restriction to add.

scripts/core/eventQueryKernelTimestamps.yml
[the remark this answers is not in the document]
joshuaranjan: @aravindksg 
I was thinking since resultsProps.pSynchronizedTimestampsBuffer is updated by the driver, when this API is called, may be this should be [out].

wdamon-intel:
@aravindksg @MichalMrozek @joshuaranjan @jandres742 Are there any remaining comments/questions/concerns?

scripts/core/EXT_EventQueryKernelTimestamps.rst
joshuaranjan: @aravindksg 
Not sure whether "kernel_timestamps_result_t" would be appropriate here, since spec mentions the below for "ze_kernel_timestamp_data_t"

_The timestamp frequency can be queried from ::ze_device_properties_t.timerResolution.
The number of valid bits in the timestamp value can be queried from ::ze_device_properties_t.kernelTimestampValidBits._

This may not be true for the synchronized timestamp.
We may have to create a new structure, unless we have something that represents host.

jandres742 (approved):

scripts/core/EXT_EventQueryKernelTimestamps.rst
wdamon-intel: @joshuaranjan 
Would it be sufficient to add a note to the details:
```
details:
    - "The timestamp frequency can be queried from $x_device_properties_t.timerResolution."
    - "The number of valid bits in the timestamp value can be queried from $x_device_properties_t.kernelTimestampValidBits."
    - "1.6": "**Note**: The `timerResolution` and `kernelTimestampValidBits` members of `$x_device_properties_t` does not apply to synchronized timestamp data"
```

aravindksg: @joshuaranjan good point. should below `std::vector<uint64_t>` be `std::vector<std::chrono::nanoseconds> synchronizedTimestamps(count)` ? @wdamon-intel

aravindksg: apologies, please disregard above comment, seeing that in latest commit this was fixed

joshuaranjan: @wdamon-intel 

> @joshuaranjan Would it be sufficient to add a note to the details:
> 
> ```
> details:
>     - "The timestamp frequency can be queried from $x_device_properties_t.timerResolution."
>     - "The number of valid bits in the timestamp value can be queried from $x_device_properties_t.kernelTimestampValidBits."
>     - "1.6": "**Note**: The `timerResolution` and `kernelTimestampValidBits` members of `$x_device_properties_t` does not apply to synchronized timestamp data"
> ```

Yes.
Also the spec mentions
_uint64_t kernelStart [out] **device clock** at start of kernel execution_

Here device clock might not apply as well.

So to handle both , if we can have something like the below, would it be OK ?
- "1.6": "**Note**: For synchronized timestamp data, the `timerResolution` and `kernelTimestampValidBits` members of `$x_device_properties_t` does not apply. Also the kernel timestamps are host timestamps

@aravindksg please share your thoughts

wdamon-intel: @joshuaranjan, @aravindksg 
Instead of all these notes, I'm thinking it'll be better to introduce new data structures: `ze_synchronized_timestamp_data_ext_t` and `ze_synchronized_timestamp_result_ext_t`

MichalMrozek (approved):

Issue #11, opened 2022-08-17
Device and Metrics timestamps synchronization


jandres742:
New API to synchronize the device and metrics timestamps. Given that source of metrics clock can vary from the main device clock it is necessary to have mechanism to synchronize time when metrics are collected. Different devices may have different metrics source clock for different metrics types, so the API should allow metric group granularity.

`ze_result_t zetDeviceGetMetricTimestamps(ze_device_handle_t hDevice, uint64_t* globalTimestamp, uint32_t *pCount,  metric_group_handle_t *phMetricGroups, uint64_t* metricTimestamps);`

jandres742:
@matcabral: one proposal is:

```cpp
ze_result_t zetDeviceGetMetricTimestamps(
ze_device_handle_t hDevice,
uint64_t* globalTimestamp,                             
uint32_t *pCount, 
metric_group_handle_t *phMetricGroups,
uint64_t* metricTimestamps);
```
do we want to have metricTimestamps and globalTimestamp synchronized to device or host?

matcabral:
@jandres742 , device.

jandres742:
@matcabral : could you check https://github.com/oneapi-src/level-zero-spec/pull/78 ?

joshuaranjan:
Hi 
Looking at the original request at https://github.com/intel-innersource/libraries.compute.oneapi.level-zero.spec/issues/34, there were 2 requirements.
1. Get a synchronized Device (or Host) , MetricGroup timestamp
2. add extended properties (in zet_metric_group_properties_t) for the metricgroup to have "timerResolution" and "timestampValidBits"

"1" is addressed here. 
Please sahre if "2" would be addreressed in another ticket ? 
I think we need both to do the mapping.

jandres742:
> Hi Looking at the original request at [intel-innersource/libraries.compute.oneapi.level-zero.spec#34](https://github.com/intel-innersource/libraries.compute.oneapi.level-zero.spec/issues/34), there were 2 requirements.
> 
> 1. Get a synchronized Device (or Host) , MetricGroup timestamp
> 2. add extended properties (in zet_metric_group_properties_t) for the metricgroup to have "timerResolution" and "timestampValidBits"
> 
> "1" is addressed here. Please sahre if "2" would be addreressed in another ticket ? I think we need both to do the mapping.

@joshuaranjan that was not mentioned explicitly, missed that. on which API do you think those should be returned?

joshuaranjan:
[...]

PR #78, opened 2023-02-22
Add extension for global metrics timestamps

jandres742:
Resolves: #11

### DIFF
--- a/scripts/tools/GlobalTimestamps.yml
+++ b/scripts/tools/GlobalTimestamps.yml
@@ -13,13 +13,13 @@ version: "1.5"
 type: macro
 desc: "Global Metric Timestamps Experimental Extension Name"
 version: "1.5"
-name: $T_GLOBAL_METRIS_TIMESTAMPS_EXP_NAME
-value: '"$T_experimental_global_metric_timestamps"'
+name: $T_GLOBAL_METRICS_TIMESTAMPS_EXP_NAME
+value: '"$XT_experimental_global_metric_timestamps"'
 --- #--------------------------------------------------------------------------
 type: enum
 desc: "Global Metric Timestamps Experimental Extension Version(s)"
 version: "1.5"
-name: $t_global_metric_timestamps_exp_version_t
+name: $x_metric_global_timestamps_exp_version_t
 etors:
     - name: "1_0"
       value: "$X_MAKE_VERSION( 1, 0 )"
@@ -29,7 +29,7 @@ type: struct
 desc: "Metric timestamps resolution"
 version: "1.5"
 class: $tMetric
-name: $x_metric_timestamps_resolution_exp_t
+name: $t_metric_global_timestamps_resolution_exp_t
 base: $x_base_desc_t
 members:
     - type: uint64_t
@@ -45,7 +45,7 @@ details:
 type: function
 desc: "Returns metric timestamps synchronized with global device timestamps, optionally synchronized with host"
 class: $tMetricGroup
-name: GetTimestampsExp
+name: GetGlobalTimestampsExp
 decl: static
 details:
     - "The application may call this function from simultaneous threads."
--- a/scripts/tools/GlobalTimestamps.yml
+++ b/scripts/tools/GlobalTimestamps.yml
@@ -58,11 +58,11 @@ params:
       name: synchronizedWithHost
       desc: "[in] Returns the timestamps synchronized to the host or the device."
       init: "false"
-    - type: "uint64_t"
+    - type: "uint64_t*"
       name: globalTimestamp
       desc: |
             [out] Device timestamp.
-    - type: "uint64_t"
+    - type: "uint64_t*"
       name: metricTimestamp
       desc: |
             [out] Metric timestamp.
--- a/scripts/tools/GlobalTimestamps.yml
+++ b/scripts/tools/GlobalTimestamps.yml
@@ -7,26 +7,43 @@
 #
 --- #--------------------------------------------------------------------------
 type: header
-desc: "Intel $OneApi Level-Zero Tool Experimental Extension for Global Metrics Timestamps"
+desc: "Intel $OneApi Level-Zero Tool Experimental Extension for Global Metric Timestamps"
 version: "1.5"
 --- #--------------------------------------------------------------------------
 type: macro
-desc: "Global Metrics Timestamps Experimental Extension Name"
+desc: "Global Metric Timestamps Experimental Extension Name"
 version: "1.5"
-name: $T_GLOBAL_METRICS_TIMESTAMPS_EXP_NAME
-value: '"$XT_experimental_global_metrics_timestamps"'
+name: $T_GLOBAL_METRIS_TIMESTAMPS_EXP_NAME
+value: '"$T_experimental_global_metric_timestamps"'
 --- #--------------------------------------------------------------------------
 type: enum
-desc: "Global Metrics Timestamps Experimental Extension Version(s)"
+desc: "Global Metric Timestamps Experimental Extension Version(s)"
 version: "1.5"
-name: $x_global_metrics_timestamps_exp_version_t
+name: $t_global_metric_timestamps_exp_version_t
 etors:
     - name: "1_0"
       value: "$X_MAKE_VERSION( 1, 0 )"
       desc: "version 1.0"
 --- #--------------------------------------------------------------------------
+type: struct
+desc: "Metric timestamps resolution"
+version: "1.5"
+class: $tMetric
+name: $x_metric_timestamps_resolution_exp_t
+base: $x_base_desc_t
+members:
+    - type: uint64_t
+      name: timerResolution
+      desc: Returns the resolution of device timer in nanoseconds used for timestamps.
+    - type: uint64_t
+      name: timestampValidBits
+      desc: "[out] Returns the number of valid bits in the timestamp value."
+details:
+    - "This structure may be returned from $tMetricGroupGetProperties via `pNext` member of $t_metric_group_properties_t"
+    - "Used for determining the total number of EUs available on device."
+--- #--------------------------------------------------------------------------
 type: function
-desc: "Returns global device timestamps synchronized with metrics timesstamps, optionally mapped to host."
+desc: "Returns metric timestamps synchronized with global device timestamps, optionally synchronized with host"
 class: $tMetricGroup
 name: GetTimestampsExp
 decl: static
@@ -37,22 +54,14 @@ params:
       name: hMetricGroup
       desc: "[in] handle of the metric group"
     - type: "$x_bool_t"
-      name: mappedToHost
-      desc: "[in] Are timestamps and metrics mapped to host, or device."
+      name: synchronizedWithHost
+      desc: "[in] Return host synchronized or device synchronized metric timestamp."
       init: "false"
-    - type: uint32_t*
-      name: pCount
+    - type: "uint64_t"
+      name: globalTimestamp
       desc: |
-            [in,out] pointer to the number of timestamps.
-            if count is zero, then the driver shall update the value with the total number of timestamps available.
-            if count is greater than the number of timestamps available, then the driver shall update the value with the correct number of timestamps.
-    - type: "uint64_t*"
-      name: globalTimestamps
+            [out] Timestamp synchronized to host or device.
+    - type: "uint64_t"
+      name: metricTimestamp
       desc: |
-            [in,out][optional][range(0, *pCount)] array of timestamp results.
-            if count is less than the number of timestamps available, then driver shall only retrieve that number of timestamps.
-    - type: "uint64_t*"
-      name: metricTimestamps
-      desc: |
-            [in,out][optional][range(0, *pCount)] array of timestamp results.
-            if count is less than the number of timestamps available, then driver shall only retrieve that number of timestamps.
+            [out] Metric timestamp.
--- a/scripts/tools/GlobalTimestamps.yml
+++ b/scripts/tools/GlobalTimestamps.yml
@@ -1,0 +1,58 @@
+#
+# Copyright (C) 2023 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+#
+# See YaML.md for syntax definition
+#
+--- #--------------------------------------------------------------------------
+type: header
+desc: "Intel $OneApi Level-Zero Tool Experimental Extension for Global Metrics Timestamps"
+version: "1.5"
+--- #--------------------------------------------------------------------------
+type: macro
+desc: "Global Metrics Timestamps Experimental Extension Name"
+version: "1.5"
+name: $T_GLOBAL_METRICS_TIMESTAMPS_EXP_NAME
+value: '"$XT_experimental_global_metrics_timestamps"'
+--- #--------------------------------------------------------------------------
+type: enum
+desc: "Global Metrics Timestamps Experimental Extension Version(s)"
+version: "1.5"
+name: $x_global_metrics_timestamps_exp_version_t
+etors:
+    - name: "1_0"
+      value: "$X_MAKE_VERSION( 1, 0 )"
+      desc: "version 1.0"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Returns global device timestamps synchronized with metrics timesstamps, optionally mapped to host."
+class: $tMetricGroup
+name: GetTimestampsExp
+decl: static
+details:
+    - "The application may call this function from simultaneous threads."
+params:
+    - type: "$t_metric_group_handle_t"
+      name: hMetricGroup
+      desc: "[in] handle of the metric group"
+    - type: "$x_bool_t"
+      name: mappedToHost
+      desc: "[in] Are timestamps and metrics mapped to host, or device."
+      init: "false"
+    - type: uint32_t*
+      name: pCount
+      desc: |
+            [in,out] pointer to the number of timestamps.
+            if count is zero, then the driver shall update the value with the total number of timestamps available.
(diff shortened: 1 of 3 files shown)


jandres742:
@matcabral, @wdamon-intel , please check.

matcabral:
@dmichaelo1 and @joshuaranjan  please take a look

scripts/tools/GlobalTimestamps.yml
```diff
@@ -0,0 +1,58 @@
...
+desc: "Global Metrics Timestamps Experimental Extension Version(s)"
+version: "1.5"
+name: $x_global_metrics_timestamps_exp_version_t
+etors:
+    - name: "1_0"
+      value: "$X_MAKE_VERSION( 1, 0 )"
+      desc: "version 1.0"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Returns global device timestamps synchronized with metrics timesstamps, optionally mapped to host."
```
matcabral, line 29: For better clarity I would invert the references. Something like 

`Returns metrics timestamps synchronized with global device timestamps`

leaving out host, since it is not in scope.

scripts/tools/GlobalTimestamps.yml
```diff
@@ -0,0 +1,58 @@
...
+class: $tMetricGroup
+name: GetTimestampsExp
```
matcabral, line 31: will the final name be `zetDeviceGetMetricTimestamps` ?

scripts/tools/GlobalTimestamps.yml
```diff
@@ -0,0 +1,58 @@
...
+decl: static
+details:
+    - "The application may call this function from simultaneous threads."
+params:
```
matcabral, line 35: for consistency with oher APIs the order of params should as 

`ze_result_t zetDeviceGetMetricTimestamps(ze_device_handle_t hDevice, uint64_t* globalTimestamp, uint32_t *pCount,  metric_group_handle_t *phMetricGroups, uint64_t* metricTimestamps);`

scripts/tools/GlobalTimestamps.yml
```diff
@@ -0,0 +1,58 @@
...
+    - type: "$t_metric_group_handle_t"
+      name: hMetricGroup
+      desc: "[in] handle of the metric group"
```
matcabral, line 38: this is an array of metric group hangles

scripts/tools/GlobalTimestamps.yml
jandres742: @matcabral : if host are not in the scope, do you want to keep or remove the mappedToHost argument I added below?

```cpp
- type: "$x_bool_t"
      name: mappedToHost
      desc: "[in] Are timestamps and metrics mapped to host, or device."
```

scripts/tools/GlobalTimestamps.yml
```diff
@@ -0,0 +1,58 @@
...
+    - type: "$x_bool_t"
+      name: mappedToHost
```
matcabral, line 40: The original request was to sync with device, which I know how to satisfy. However, its unclear if synch with host can be done. Does L0 core have APIs to synch device and host? if so, its internal implementation could be used here.

scripts/tools/GlobalTimestamps.yml
matcabral: regarding host support, I saw below the boolean flag added.

scripts/tools/GlobalTimestamps.yml
```diff
@@ -0,0 +1,58 @@
...
+      init: "false"
+    - type: uint32_t*
+      name: pCount
+      desc: |
+            [in,out] pointer to the number of timestamps.
+            if count is zero, then the driver shall update the value with the total number of timestamps available.
+            if count is greater than the number of timestamps available, then the driver shall update the value with the correct number of timestamps.
+    - type: "uint64_t*"
+      name: globalTimestamps
+      desc: |
```
matcabral, line 51: global timestamp is only one value. Not an array.

scripts/tools/GlobalTimestamps.yml
```diff
@@ -0,0 +1,58 @@
...
+            [in,out][optional][range(0, *pCount)] array of timestamp results.
+            if count is less than the number of timestamps available, then driver shall only retrieve that number of timestamps.
+    - type: "uint64_t*"
+      name: metricTimestamps
+      desc: |
+            [in,out][optional][range(0, *pCount)] array of timestamp results.
```
matcabral, line 57: `array of timestamp results, one per metric group`

scripts/tools/GlobalTimestamps.yml
joshuaranjan, line 29: "optionally mapped to host" ->"optionally synchronized with host" ?

scripts/tools/GlobalTimestamps.yml
joshuaranjan: "mappedToHost" to  "SynchronizedWithHost" ?

scripts/tools/GlobalTimestamps.yml
jandres742: @matcabral : could you clarify, whether we want then to have host mapped support or not?

scripts/tools/GlobalTimestamps.yml
jandres742: @matcabral : that's one option. if you have another option, could you propose?

scripts/tools/GlobalTimestamps.yml
```diff
@@ -0,0 +1,58 @@
...
+decl: static
+details:
+    - "The application may call this function from simultaneous threads."
+params:
+    - type: "$t_metric_group_handle_t"
+      name: hMetricGroup
+      desc: "[in] handle of the metric group"
+    - type: "$x_bool_t"
+      name: mappedToHost
+      desc: "[in] Are timestamps and metrics mapped to host, or device."
```
joshuaranjan, line 41: Can we rename like below:
"Are timestamps and metrics mapped to host, or device." - Return Host Synchronized or Device Synchronized metric timestamp" ?

scripts/tools/GlobalTimestamps.yml
```diff
@@ -0,0 +1,58 @@
...
+      desc: |
+            [in,out] pointer to the number of timestamps.
+            if count is zero, then the driver shall update the value with the total number of timestamps available.
+            if count is greater than the number of timestamps available, then the driver shall update the value with the correct number of timestamps.
+    - type: "uint64_t*"
+      name: globalTimestamps
+      desc: |
+            [in,out][optional][range(0, *pCount)] array of timestamp results.
+            if count is less than the number of timestamps available, then driver shall only retrieve that number of timestamps.
+    - type: "uint64_t*"
```
joshuaranjan, line 54: Since this API takes metric group handle as parameter, metric timestamp would also be one ?

scripts/tools/GlobalTimestamps.yml
jandres742: @matcabral isn't the idea to query for available metric groups with zetMetricGroupGet, where we are already returning an array, and then operate on which metric group individually?

scripts/tools/GlobalTimestamps.yml
jandres742: @matcabral : Core has something similar, but dont know if it could be exactly be used. This is an experimental extension, so if we are not suer, we could add the host support later, or remove it. Please let us know which you prefer.

scripts/tools/GlobalTimestamps.yml
jandres742: ok

scripts/tools/GlobalTimestamps.yml
jandres742: Please see above.

jandres742:
@matcabral , @joshuaranjan : 

could you confirm two things:
- is the mapping to host needed or not?
- why do we want to return all timestamps for all metrics groups, instead of one individually?

jandres742:
@matcabral , @joshuaranjan please see update.
Added the timerresolution and valid timestamp bits. Those are returned by [zetMetricGroupGetProperties](https://spec.oneapi.io/level-zero/latest/tools/api.html?highlight=metricgroupgetproperties#zetmetricgroupgetproperties), which takes a metric group handle, so that's why the new extension accepts also a metric group handle. if we were to pass an array of metric group handles, then we would need to have also an array of timestamp valid bits and resolution returned somewhere, so I think it is better to have per metric group.

for now, leaving the boolean to decide whether we want to sync to host or device, but we can remove it if wanted.

please let me know what you think

scripts/tools/GlobalTimestamps.yml
```diff
@@ -37,22 +54,14 @@ params:
       name: hMetricGroup
       desc: "[in] handle of the metric group"
     - type: "$x_bool_t"
-      name: mappedToHost
-      desc: "[in] Are timestamps and metrics mapped to host, or device."
+      name: synchronizedWithHost
+      desc: "[in] Return host synchronized or device synchronized metric timestamp."
```
joshuaranjan, line 58: Can we mention "Returns synchronized Host (or device) and metric timestamps."

This would be similar to the description for "zeDeviceGetGlobalTimestamps", which mentions "Returns synchronized Host and device global timestamps."

scripts/tools/GlobalTimestamps.yml
matcabral: that name is good, my point was about how yaml woudl concatenate strings.

scripts/tools/GlobalTimestamps.yml
matcabral: found that MDAPI also has support to synch with host. Therefore, I think makes more sense to have in on the initial implemenation .

scripts/tools/GlobalTimestamps.yml
matcabral: yes, lets have it please. Confirmed also in comment below.

scripts/tools/GlobalTimestamps.yml
matcabral: correct for metric groups handles collection with zetMetricGroupGet(). However, the idea to support multiple metric group handles and timestamps in the synch API is that different metric groups could have different clock sources. Therefore, user may whant to confirm they are all in synch

matcabral:
> Just my opinion: Considering that we can capture metric for domain independent metric groups at a single time, I dont think we need to return for multiple metric groups. 

@joshuaranjan , But the user would need to query each metric group, find the the domain and then compare domains to find which metric groups to submit ? In this case would make more sense for this API to accept metric domain as an input paramters. 

>Added the timerresolution and valid timestamp bits. 

Following the above, would also asume these will be by domain, not metric group. 

>for now, leaving the boolean to decide whether we want to sync to host or device, but we can remove it if wanted.

yes please, makes sense to have it.

scripts/tools/GlobalTimestamps.yml
jandres742: @matcabral  final name would be zetMetricGroupGetTimestampsExp

scripts/tools/GlobalTimestamps.yml
```diff
@@ -7,26 +7,43 @@
...
+members:
+    - type: uint64_t
+      name: timerResolution
+      desc: Returns the resolution of device timer in nanoseconds used for timestamps.
+    - type: uint64_t
+      name: timestampValidBits
+      desc: "[out] Returns the number of valid bits in the timestamp value."
+details:
+    - "This structure may be returned from $tMetricGroupGetProperties via `pNext` member of $t_metric_group_properties_t"
+    - "Used for determining the total number of EUs available on device."
```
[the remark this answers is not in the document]
jandres742: oh yes, copy paste error :)

scripts/tools/GlobalTimestamps.yml
jandres742: @joshuaranjan here we are not returning both, only host or device sync'd.

jandres742:
> > Just my opinion: Considering that we can capture metric for domain independent metric groups at a single time, I dont think we need to return for multiple metric groups.
> 
> @joshuaranjan , But the user would need to query each metric group, find the the domain and then compare domains to find which metric groups to submit ? In this case would make more sense for this API to accept metric domain as an input paramters.
> 
> > Added the timerresolution and valid timestamp bits.
> 
> Following the above, would also asume these will be by domain, not metric group.
> 
> > for now, leaving the boolean to decide whether we want to sync to host or device, but we can remove it if wanted.
> 
> yes please, makes sense to have it.

@matcabral : by metric domain I assume you mean device. 

now, if we return an array for all metric groups in the same API, would the order of the timestamps array be the same as the order of metric groups returned by zetMetricGroupGet? What happens if we return the array of timestamps, supposedly for all metric groups, and a metric group cannot support this API, should we return a value (0 or -1) to keep the same order?

scripts/tools/GlobalTimestamps.yml
joshuaranjan: @jandres742 , Ok.

>Return host synchronized or device synchronized metric timestamp

This seemed to say that there is only one output value from this API i.e the "metric timestamp"

>"Returns synchronized Host (or device) and metric timestamps.

This seemed to say that there are two output values from this API 
1. the "metric timestamp"
2. Device OR host timestamp

May be this is subjective to my perception. Kiindly use which is better.
Thanks

joshuaranjan:
@jandres742 
Metric domain number defines which set of metrics could be collected concurrently.
Currently there are several metric groups (contains a specific set of metrics) which are collected from OA.
All these metric groups share the same domain number.
This means that there is a single set of metrics (1 metric group) that could be collected at a single instance.

For example: Stall Sampling Metric Group has a different domain number since it is not collected using OA.
So Stall Sampling Metric group could be collected concurrently with an OA based metric group.

jandres742:
> @jandres742 Metric domain number defines which set of metrics could be collected concurrently. Currently there are several metric groups (contains a specific set of metrics) which are collected from OA. All these metric groups share the same domain number. This means that there is a single set of metrics (1 metric group) that could be collected at a single instance.
> 
> For example: Stall Sampling Metric Group has a different domain number since it is not collected using OA. So Stall Sampling Metric group could be collected concurrently with an OA based metric group.

thanks @joshuaranjan . So do we want to have device being passed here then?

joshuaranjan:
>thanks @joshuaranjan . So do we want to have device being passed here then?

Not necessarily. (I assume "device" here means ze_device_handle_t)
Since domain is not attached to "device".
Please note that, a device can have multiple metric domain numbers supported.

Metric groups are already device specific ([zetMetricGroupGet](https://spec.oneapi.io/level-zero/latest/tools/api.html?highlight=zetmetricgroup#zetmetricgroupget))
I still think we could query based on metric group handle. (Please check reasoning here: https://github.com/oneapi-src/level-zero-spec/pull/78#issuecomment-1445723141)

Thanks

joshuaranjan:
>thanks @joshuaranjan . So do we want to have device being passed here then?

Not necessarily. (I assume "device" here means ze_device_handle_t)
Since domain is not attached to "device".
Please note that, a device can have multiple metric domain numbers supported.

Metric groups are already device specific ([zetMetricGroupGet](https://spec.oneapi.io/level-zero/latest/tools/api.html?highlight=zetmetricgroup#zetmetricgroupget))
I still think we could query based on metric group handle. (Please check reasoning here: https://github.com/oneapi-src/level-zero-spec/pull/78#issuecomment-1445723141)

Thanks

matcabral:
Sumarizing few above topics:
- Make the API accept only one metric group per call. Rationale: make it simple on the first iteration. If needed we extend it to an array while the extension is still experimental. 
- Use metric group instead of metric domain. Rationale: user simplicity. 
- Use metric_timestamps_resolution_exp_t instead of just timestamp. Rationale: includes valid bits. 
- Use boolean to choose from gloal and host timesamp against wihch to normalize. Rationale: more flexibilty to user.

matcabral:
>@matcabral : by metric domain I assume you mean device.

@jandres742 , not really, this is part of the properties of each metric group. As explained above tells what metric groups can be collected concurrently. 
https://spec.oneapi.io/level-zero/latest/tools/api.html?highlight=zetmetricgroup#_CPPv429zet_metric_group_properties_t

uint32_t domain
[out] metric group domain number. Cannot use multiple, simultaneous metric groups from the same domain.

scripts/tools/GlobalTimestamps.yml
matcabral: @jandres742 would you please share how the signature of the API will look like with the latest changes? I'm not seeing the device hangle as input or the metric group timestamp .

joshuaranjan: Based on my "assumption" of how this .yml will be processed, it might look like below

```
zetGetGlobalTimestampsExp(
                    zet_metric_group_handle_t hMetricGroup, 
                    ze_bool_t synchronizedWithHost, 
                    uint64_t* globalTimestamp, 
                    uint64_t* metricTimestamp)
```
I assume the return type would be **ze_result_t**

@jandres742 could confirm

matcabral: why are we adding _metric_global_timestamps_resolution_exp_t ?

joshuaranjan: >why are we adding _metric_global_timestamps_resolution_exp_t ?

We need to know the timer frequency and when the timer will overflow (valid bits).
These will be used for mapping the timestamp to other timers (host clock)

matcabral: perfect!

joshuaranjan (approved):

matcabral (approved):

Issue #61, opened 2023-01-13
Level0 APi which would return a string given ze_result_t error code


RaviNarayanaswamy:
We would like to print a meaningful error when level returns an error code
eg:- 

```cpp
ze_result_t ret = zeCommandListAppendLaunchKernel()
if (ret != ZE_RESULT_SUCCESS) {
    zeDriverGetErrorString(&str);
    printf("Error returned by L0: %s\n", str);
}
```

Currently we need to do this
```
  export PrintDebugMessages=1
  export NEOReadDebugKeys=1
```

jandres742:
@zackwaters , @wdamon-intel : what do you think about this?

wdamon-intel:
> [ZW, WD] what do you think about this?

I'm not opposed to idea, but it should be an SDK enhancement over a Core API feature.

kevin-harms:
Why wouldn't it be added to the Core API?

zackwaters:
I don't have any objection to this being added to the Core API.

jandres742:
@wdamon-intel : could you elaborate on your suggestion above?

wdamon-intel:
Basically, when I read this request, I saw an immediate parallel to the `vk_enum_string_helper.h` that is part of the Vulkan SDK. Certainly this would be a useful developer/debug tool, but does it merit a Core API?

That was my thinking... I'm not strictly opposed to it being a part of the Core API, however.

jandres742:
thanks @wdamon-intel .
So use case here is that when doing something, like appending a kernel, L0 implementation finds there's a problem, for instance, SLM size is not enough. L0 driver would then return an error, maybe ZE_RESULT_ERROR_INVALID_ARGUMENT. However, that's not descriptive enough for the application to know what happened. With the interface suggested here, additional information could be returned.

Now, another solution of course would be to add a new error code for that, like in the example I mentioned above, we could have ZE_RESULT_ERROR_SLM_SIZE, or something similar. This means L0 spec would need to define more codes, which also seems to be the approach taken by Vulkan based on what is in https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/layers/generated/vk_enum_string_helper.h, since there's a lot of  error codes there :)

So we have I think two options:

1. Add errors codes to cover any new and possible scenario that could cause an L0 call to fail.
or
[...]

PR #90, opened 2023-03-06
Add zeDriverGetLastResultString

jandres742:
This function retrieves a message associated with the returned code of the last call made.

Resolves: #61

### DIFF
--- a/scripts/core/driver.yml
+++ b/scripts/core/driver.yml
@@ -273,15 +273,15 @@ members:
       desc: "[in] handle of the driver instance"
 --- #--------------------------------------------------------------------------
 type: function
-desc: "Retrieves a message associated with the last error code returned by the driver in the current thread."
+desc: "Retrieves a string describing the last error code returned by the driver in the current thread."
 version: "1.6"
 class: $xDriver
-name: GetLastErrorString
+name: GetLastErrorDescription
 details:
     - "String returned is thread local."
     - "String is only updated on calls returning an error, i.e., not on calls returning $X_RESULT_SUCCESS."
     - "String may be empty if driver considers error code is already explicit enough to describe cause."
-    - "Memory pointed to by ppString is owned by driver."
+    - "Memory pointed to by ppString is owned by the driver."
     - "String returned is null-terminated."
 params:
     - type: $x_driver_handle_t
@@ -289,4 +289,4 @@ params:
       desc: "[in] handle of the driver instance"
     - type: "const char**"
       name: ppString
-      desc: "[in,out] pointer to an null-terminated array of characters containing message associated with the last call."
+      desc: "[in,out] pointer to a null-terminated array of characters describing cause of error."
--- a/scripts/core/driver.yml
+++ b/scripts/core/driver.yml
@@ -291,6 +291,3 @@ params:
     - type: "const char**"
       name: ppString
       desc: "[in,out] pointer to an array of characters containing message associated with the last call."
-    - type: "size_t*"
-      name: pSize
-      desc: "[in,out][optional] size of message string pointed by ppString."
--- a/scripts/core/driver.yml
+++ b/scripts/core/driver.yml
@@ -271,3 +271,26 @@ members:
     - type: $x_driver_handle_t
       name: handle
       desc: "[in] handle of the driver instance"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Retrieves a message associated with the returned code of the last call made."
+version: "1.6"
+class: $xDriver
+name: GetLastResultString
+details:
+    - "String returned is thread local."
+    - |
+       String may be empty if last call returned $X_RESULT_SUCCESS or if driver
+       determines that error code is already explicit enough to convey error cause.
+    - "Memory pointed by ppString is owned by driver."
+    - "String returned is null-terminated."
+params:
+    - type: $x_driver_handle_t
+      name: hDriver
+      desc: "[in] handle of the driver instance"
+    - type: "const char**"
+      name: ppString
+      desc: "[in,out] pointer to an array of characters containing message associated with the last call."
+    - type: "size_t*"
+      name: pSize
+      desc: "[in,out][optional] size of message string pointed by ppString."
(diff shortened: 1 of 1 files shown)


scripts/core/driver.yml
```diff
@@ -271,3 +271,26 @@ members:
...
+    - "String returned is null-terminated."
+params:
+    - type: $x_driver_handle_t
+      name: hDriver
+      desc: "[in] handle of the driver instance"
+    - type: "const char**"
+      name: ppString
+      desc: "[in,out] pointer to an array of characters containing message associated with the last call."
+    - type: "size_t*"
+      name: pSize
```
bmyates, line 295: `pSize` is unnecessary because string is null-terminated

jandres742: @bmyates . right, I was thinking on that. Users could just use strlen.

scripts/core/driver.yml
```diff
@@ -271,3 +271,23 @@ members:
     - type: $x_driver_handle_t
       name: handle
       desc: "[in] handle of the driver instance"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Retrieves a message associated with the returned code of the last call made."
+version: "1.6"
+class: $xDriver
+name: GetLastResultString
```
MichalMrozek, line 279: what's the goal of this function ?
how would this work in the driver?

scripts/core/driver.yml
```diff
@@ -271,3 +271,23 @@ members:
...
+       determines that error code is already explicit enough to convey error cause.
+    - "Memory pointed by ppString is owned by driver."
+    - "String returned is null-terminated."
+params:
+    - type: $x_driver_handle_t
+      name: hDriver
+      desc: "[in] handle of the driver instance"
+    - type: "const char**"
+      name: ppString
+      desc: "[in,out] pointer to an array of characters containing message associated with the last call."
```
wdamon-intel, line 293: Is this array of characters null-terminated? If yes, we should explicitly state that here.

scripts/core/driver.yml
wdamon-intel: Maybe a better name would be `GetLastResultDescription`?

MichalMrozek: Yeah I wonder how the string should be stored in the driver.
This would need to be dynamic memory allocation (i.e. string)
What about thread safety, if multiple calls are done in parallel ?
Does successful call free the string buffer?

After the API call, the string buffer needs to be still allocated or it can be freed?

wdamon-intel: Would there be any opposition to these simply being static strings, similar to the `string_Vk...` routines in Vulkan SDK's `vk_enum_string_helper.h`?

jandres742: @MichalMrozek : interface is defined as thread-local, so this is expected to return a string describing the last returned code by that thread.

Right, I was thinking we would have a thread-local static array, and whenever the driver returns an error, like ZE_RESULT_UNKNOWN, driver will add a message to that static array. And on the next call, made by that that thread, the array is either cleaned (if we are returning ZE_RESULT_SUCCESS) or rewritten if we are returning a new error code.

To avoid having to write the array on SUCCESS calls, maybe we could add the provision that this returns a message associated with the last error code, and that message stays in the array until the next error code:

Calls that return SUCCESS dont overwrite the array or message.

scripts/core/driver.yml
jandres742: it is mentioned in the details, but we can add it here as well of course.

scripts/core/driver.yml
MichalMrozek: I would like to make sure that driver does not need to do anything on SUCCESS case.
SUCCESS case needs to be razor thin efficient.

jandres742: @MichalMrozek : right, that's a very good point and that aligns with expectation, since this is more for errors than for success. So @wdamon-intel  suggestion of calling it "GetLastErrorDescription" is good, and we could add something like:

```html
- String is updated only when driver returns an error code.
- Calling zeDriverGetLastError after a call that returns SUCCESS will retrieve the same string associated with last call that returned an error other than SUCCESS.
```

scripts/core/driver.yml
jandres742: Added.

scripts/core/driver.yml
```diff
@@ -271,3 +271,23 @@ members:
...
+desc: "Retrieves a message associated with the returned code of the last call made."
+version: "1.6"
+class: $xDriver
+name: GetLastResultString
+details:
+    - "String returned is thread local."
+    - |
+       String may be empty if last call returned $X_RESULT_SUCCESS or if driver
+       determines that error code is already explicit enough to convey error cause.
+    - "Memory pointed by ppString is owned by driver."
```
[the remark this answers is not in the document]
jandres742: Added.

scripts/core/driver.yml
jandres742: addressed comments

scripts/core/driver.yml
```diff
@@ -271,3 +271,22 @@ members:
...
+    - "String may be empty if driver considers error code is already explicit enough to describe cause."
+    - "Memory pointed to by ppString is owned by driver."
+    - "String returned is null-terminated."
+params:
+    - type: $x_driver_handle_t
+      name: hDriver
+      desc: "[in] handle of the driver instance"
+    - type: "const char**"
+      name: ppString
+      desc: "[in,out] pointer to an null-terminated array of characters containing message associated with the last call."
```
wdamon-intel, line 292: "pointer to **_a_** null-terminated..."

scripts/core/driver.yml
```diff
@@ -271,3 +271,22 @@ members:
...
+type: function
+desc: "Retrieves a message associated with the last error code returned by the driver in the current thread."
+version: "1.6"
+class: $xDriver
+name: GetLastErrorString
+details:
+    - "String returned is thread local."
+    - "String is only updated on calls returning an error, i.e., not on calls returning $X_RESULT_SUCCESS."
+    - "String may be empty if driver considers error code is already explicit enough to describe cause."
+    - "Memory pointed to by ppString is owned by driver."
```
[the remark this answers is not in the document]
jandres742: added.

wdamon-intel (approved):
Looks good to me.

MichalMrozek (commented):
lgtm

Issue #94, opened 2023-03-09
[Metrics] API to perform offline processing of metrics raw data


joshuaranjan:
Request is to enable the following
1. Machine independent metrics calculation - Metrics calculation could be done on a machine which is different from where it was collected (using "Read")
2. Applications / Clients would want to process the raw metrics without using L0 interface. So enough information about how to process the raw metrics could be provided.

joshuaranjan:
Hi @jandres742 ,
If possible please change to "API:Metrics" tag instead of "Sysman" ?
Thanks

joshuaranjan:
@matcabral @dmichaelo
Please check this

joshuaranjan:
PR : https://github.com/oneapi-src/level-zero-spec/pull/95

PR #95, opened 2023-03-09
Add support for machine independent calculation for metrics data

joshuaranjan:
Resolves: #94

### DIFF
--- a/scripts/tools/EXT_Exp_MetricExportData.rst
+++ b/scripts/tools/EXT_Exp_MetricExportData.rst
@@ -1,0 +1,103 @@
+<%
+import re
+from templates import helper as th
+%><%
+    OneApi=tags['$OneApi']
+    x=tags['$x']
+    X=x.upper()
+    t=tags['$t']
+    T=t.upper()
+%>
+:orphan:
+
+.. _ZET_experimental_metric_export_data:
+
+==========================================
+Metrics Export Data Experimental Extension
+==========================================
+
+API
+----
+
+* Structures
+
+    * ${t}_metric_export_data_format_exp_t
+    * ${t}_metric_calculate_exp_desc_t
+
+* Functions
+
+    * ${t}MetricGroupGetExportDataExp
+    * ${x}DriverCalculateMetricExportDataExp
+
+Sample Code
+------------
+
+The following code shows how to export metrics raw data.
+
+.. parsed-literal::
+
+    // Get size for export data and count for data format
+    size_t exportDataSize = 0;
+    uint32_t exportDataFormatCount = 0;
+    ${t}MetricGroupGetExportDataExp (hMetricGroup, rawData, rawDataSize,
+                                &exportDataSize, &exportDataFormatCount, nullptr, nullptr);
+
+    // Allocate for export data and data format
+    uint8_t* pExportData = malloc(exportDataSize);
+    ${t}_metric_export_data_format_exp_t* pExportFormatData =
+                            malloc(sizeof(${t}_metric_export_data_format_exp_t) * exportDataFormatCount);
+
+    // Retrieve export data and data format
+    ${t}MetricGroupGetExportDataExp (hMetricGroup, rawData, rawDataSize,
+                        &exportDataSize, &exportDataFormatCount, pExportData, pExportFormatData);
+
+
+
+The following code shows how to perform machine independent calculation using the exported data.
+
+.. parsed-literal::
+
+    // Get data count and total metric count
+    uint32_t dataCount = 0;
+    uint32_t totalMetricCount = 0;
+    ${t}_metric_group_calculation_type_t type = ${T}_METRIC_GROUP_CALCULATION_TYPE_METRIC_VALUES;
+    ${t}_metric_calculate_exp_desc_t calculateDesc;
+    calculateDesc.rawReportSkipCount = 0;
+    ${x}DriverCalculateMetricExportDataExp(hDriver, type, exportDataSize, pExportData, &calculateDesc,
+                        &dataCount, &totalMetricCount, nullptr, nullptr);
+
+    void* pMetricCounts = malloc(dataCount * sizeof(uint32_t));
+    void* pMetricValues = malloc(totalMetricCount * sizeof(zet_typed_value_t));
+
+    // Get metric counts and metric values
+    ${x}DriverCalculateMetricExportDataExp(hDriver, type, exportDataSize, pExportData, &calculateDesc,
+                        &dataCount, &totalMetricCount, pMetricCounts, pMetricValues);
+
+
+The following code shows how to optionally process exported data format elements.
+
+.. parsed-literal::
+
+    // Optional: Processing of Data format elements
+    while(exportDataFormatCount--){
+
+    if(strcmp(pDataFormat->name, "GlobalSymbolName") == 0){
+        // Read Global Symbol Name
+        const char* symbolName = (const char*)pExportData;
+
+        // Do appropriate handling for "GlobalSymbolName"
+    }
+
+    if(strcmp(pDataFormat->name, "GlobalSymbolValue") == 0){
+        // Read Global Symbol Value
+        if(pDataFormat->type == ZET_VALUE_TYPE_UINT16){
+            uint16_t value = 0;
+            memcpy(&value, pExportData, sizeof(uint16_t);
+        }
+
+        // Do appropriate handling for "GlobalSymbolValue"
+    }
+
+    pExportData += pDataFormat->lengthInBytes;
+    pDataFormat++;
+    }
--- a/scripts/tools/EXT_Exp_MetricExportData.rst
+++ b/scripts/tools/EXT_Exp_MetricExportData.rst
@@ -21,7 +21,6 @@ API
 
 * Structures
 
-    * ${t}_metric_export_data_format_exp_t
     * ${t}_metric_calculate_exp_desc_t
 
 * Functions
@@ -32,25 +31,21 @@ API
 Sample Code
 ------------
 
-The following code shows how to export metrics raw data.
+The following code shows how to export metrics raw data for a metric group.
 
 .. parsed-literal::
 
-    // Get size for export data and count for data format
+    // Get size for export data
     size_t exportDataSize = 0;
-    uint32_t exportDataFormatCount = 0;
     ${t}MetricGroupGetExportDataExp (hMetricGroup, rawData, rawDataSize,
-                                &exportDataSize, &exportDataFormatCount, nullptr, nullptr);
+                                    &exportDataSize, nullptr);
 
-    // Allocate buffer for export data and data format
+    // Allocate buffer for export data
     uint8_t* pExportData = malloc(exportDataSize);
-    ${t}_metric_export_data_format_exp_t* pExportFormatData =
-                            malloc(sizeof(${t}_metric_export_data_format_exp_t) * exportDataFormatCount);
 
     // Retrieve export data and data format
     ${t}MetricGroupGetExportDataExp (hMetricGroup, rawData, rawDataSize,
-                        &exportDataSize, &exportDataFormatCount, pExportData, pExportFormatData);
-
+                                    &exportDataSize, pExportData);
 
 
 The following code shows how to perform metrics calculation of collected data, which can be done in a different system than where data was collected.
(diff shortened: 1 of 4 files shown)


scripts/tools/EXT_Exp_MetricExportData.rst
```diff
@@ -0,0 +1,99 @@
...
+    X=x.upper()
+    t=tags['$t']
+    T=t.upper()
+%>
+:orphan:
+
+.. _ZET_experimental_metric_export_data:
+
+==========================================
+Metrics Export Data Extension
```
jandres742, line 16: Data Experimental Extension

scripts/tools/metricExportData.yml
```diff
@@ -0,0 +1,159 @@
...
+type: struct
+desc: "Metrics calculation descriptor"
+version: "1.6"
+class: $tMetrics
+name: $t_metric_calculate_exp_desc_t
+base: $t_base_desc_t
+members:
+    - type: uint32_t
+      name: "rawReportSkipCount"
+      desc: "[in] number of reports to skip during calculation
```
jandres742, line 70: missing quote

scripts/tools/EXT_Exp_MetricExportData.rst
joshuaranjan: updated

scripts/tools/metricExportData.yml
joshuaranjan: Fixed it.
Thanks

scripts/tools/EXT_Exp_MetricExportData.rst
```diff
@@ -0,0 +1,103 @@
...
+
+.. parsed-literal::
+
+    // Get size for export data and count for data format
+    size_t exportDataSize = 0;
+    uint32_t exportDataFormatCount = 0;
+    ${t}MetricGroupGetExportDataExp (hMetricGroup, rawData, rawDataSize,
+                                &exportDataSize, &exportDataFormatCount, nullptr, nullptr);
+
+    // Allocate for export data and data format
```
matcabral, line 45: minor:  `Allocate buffer ..`

scripts/tools/EXT_Exp_MetricExportData.rst
```diff
@@ -0,0 +1,103 @@
...
+    ${t}_metric_export_data_format_exp_t* pExportFormatData =
+                            malloc(sizeof(${t}_metric_export_data_format_exp_t) * exportDataFormatCount);
+
+    // Retrieve export data and data format
+    ${t}MetricGroupGetExportDataExp (hMetricGroup, rawData, rawDataSize,
+                        &exportDataSize, &exportDataFormatCount, pExportData, pExportFormatData);
+
+
+
+The following code shows how to perform machine independent calculation using the exported data.
```
matcabral, line 56: `machine independent` a bit ambiguous. How about saying something like ` .. how to perform metrics calculation of collected data, which can be done in a different system` ?

scripts/tools/metricExportData.yml
```diff
@@ -0,0 +1,159 @@
...
+            [in,out][optional][range(0, *pExportDataFormatCount)] buffer of exported data format
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Calculate one or more sets of metric values from exported raw data."
+version: "1.6"
+class: $xDriver
+name: CalculateMetricExportDataExp
+decl: static
+details:
+    - "This function is similar to $tMetricGroupCalculateMultipleMetricValuesExp except it would calculate from exported metric data."
```
matcabral, line 120: IMHO would start by describing what the API does, then add the relations. That seems clearer. 

`Calculate metrics values with raw data returned by $GetExportDataExp, which can be done in different node that where data was collected. This function is similar to $tMetricGroupCalculateMultipleMetricValuesExp except it calculates from exported metric data.`

scripts/tools/metricExportData.yml
```diff
@@ -0,0 +1,159 @@
...
+      desc: "[in] number of reports to skip during calculation"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Export Metrics Data for machine independent calculation."
+version: "1.6"
+class: $tMetricGroup
+name: GetExportDataExp
+decl: static
+details:
+    - "This function exports or stores raw data and necessary information to perform machine independent calculation."
```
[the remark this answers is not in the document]
matcabral: ` don't need to have GPUs` 
we can't say GPUs, maybe accelerator.

scripts/tools/metricExportData.yml
joshuaranjan: Thanks. Please check the updated one.

scripts/tools/metricExportData.yml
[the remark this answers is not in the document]
joshuaranjan: Sure .. Modified for it

scripts/tools/EXT_Exp_MetricExportData.rst
joshuaranjan: Ok . Modified

scripts/tools/EXT_Exp_MetricExportData.rst
joshuaranjan: Updated. Thanks

scripts/tools/EXT_Exp_MetricExportData.rst
```diff
@@ -53,7 +53,7 @@ The following code shows how to export metrics raw data.
 
 
 
-The following code shows how to perform machine independent calculation using the exported data.
+The following code shows how to perform metrics calculation of collected data, which can be done in a different system.
```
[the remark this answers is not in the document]
joshuaranjan: Updated

scripts/tools/metricExportData.yml
```diff
@@ -70,13 +70,13 @@ members:
...
 type: function
-desc: "Export Metrics Data for machine independent calculation."
+desc: "Export Metrics Data for system independent calculation."
 version: "1.6"
 class: $tMetricGroup
 name: GetExportDataExp
 decl: static
 details:
-    - "This function exports or stores raw data and necessary information to perform machine independent calculation."
+    - "This function exports raw data and necessary information to perform metrics calculation of collected data in a different system which may or may not have accelerators."
```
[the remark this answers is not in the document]
joshuaranjan: updated

matcabral (approved):

scripts/tools/common.yml
```diff
@@ -108,6 +108,9 @@ etors:
     - name: TRACER_EXP_DESC
       desc: $t_tracer_exp_desc_t
       value: "0x00010001"
+    - name: METRICS_CALCULATE_DESC_EXP
```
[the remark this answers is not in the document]
joshuaranjan: Fixed. Thanks

jfedorov:
> This approach is similar to the existing API design of choosing a metric to collect on a platform. Since metric(name, description) are platform specific, the user calls a zetMetricGroupGet, reads its properties and does a string-compare using the metric name or its properties to choose their interested metric for collection.

Since derived metric calculation from raw metrics is also platform specific, they are represented as strings with description so that another vendor implementatation could choose to use a different data format for representing the calculation. So here we do not specify the format explicitly.
<
Yes. Agree on the metrics as char strings. 
The concern here is Math Operations (to be used in Equations) are not Platform specific. 
The Equation types: MetricCalculationEquation, ReadEquations are also to be specified. 
Let's make an API as less obscure as possible,

scripts/tools/metricExportData.yml
```diff
@@ -0,0 +1,159 @@
...
+value: "256"
+--- #--------------------------------------------------------------------------
+type: struct
+desc: "Export Data format"
+version: "1.6"
+class: $tMetric
+name: $t_metric_export_data_format_exp_t
+members:
+    - type: char
+      name: "name[$T_MAX_METRIC_EXPORT_DATA_ELEMENT_NAME_EXP]"
```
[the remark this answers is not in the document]
joshuaranjan: Thanks for the feedback.
To allow the data format to be strongly typed, an implementation specific header file would be used to describe it.
This would allow the exported data access to be easier to read and quicker to process.
Also this header file could support backward compatibility of data formats and version reporting for maintainability and error reporting

joshuaranjan:
@jfedorov @matcabral 
Could you please review and approve if the L0 spec changes are OK.
Thanks!

scripts/tools/metricExportData.yml
```diff
@@ -0,0 +1,129 @@
...
+    - type: size_t
+      name: exportDataSize
+      desc: "[in] size in bytes of exported data buffer"
+    - type: "const uint8_t*"
+      name: pExportData
+      desc: "[in][range(0, exportDataSize)] buffer of exported data to calculate"
+    - type: "$t_metric_calculate_exp_desc_t*"
+      name: pCalculateDescriptor
+      desc: "[in] descriptor specifying calculation specific parameters"
+    - type: uint32_t*
```
jandres742, line 109: this API has two pCount parameters, so, should we add some of the same clarifications we have for https://github.com/oneapi-src/level-zero-spec/pull/96 ?

scripts/tools/metricExportData.yml
```diff
@@ -0,0 +1,129 @@
...
+      name: pSetCount
```
jandres742, line 110: these two pCount are not marked as optional, so they must be passed always, at the same time?

scripts/tools/EXT_Exp_MetricExportData.rst
```diff
@@ -0,0 +1,103 @@
...
+
+* Structures
+
+    * ${t}_metric_export_data_format_exp_t
+    * ${t}_metric_calculate_exp_desc_t
+
+* Functions
+
+    * ${t}MetricGroupGetExportDataExp
+    * ${x}DriverCalculateMetricExportDataExp
```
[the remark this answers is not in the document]
joshuaranjan: We are passing a driver handle as the first parameter, so considered ze as the prefix
It would be better to have zet. 
But would it be OK to have zet prefix and not have MetricHandles passed ?

scripts/tools/metricExportData.yml
joshuaranjan: Yes.. 
We need to query for the 
* number of sets
* total number of metrics available

So that application can allocate accordingly

scripts/tools/metricExportData.yml
joshuaranjan: Would it be OK if we could add changes from #96 after it is merged, since the objective of #96 is different from this one ?

scripts/tools/metricExportData.yml
```diff
@@ -0,0 +1,129 @@
...
+type: macro
+desc: "Exporting Metrics Data Experimental Extension Name"
+version: "1.6"
+name: $T_EXPORT_METRICS_DATA_EXP_NAME
+value: '"$XT_experimental_metric_export_data"'
+--- #--------------------------------------------------------------------------
+type: enum
+desc: "Exporting Metrics Data Experimental Extension Version(s)"
+version: "1.6"
+name: $x_export_metric_data_exp_version_t
```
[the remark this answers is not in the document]
joshuaranjan: fixed. Thanks

jfedorov:
thank you. I approve it on behalf of VTune team

scripts/tools/EXT_Exp_MetricExportData.rst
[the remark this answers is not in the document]
joshuaranjan: Hi @jandres742 ,

>We are passing a driver handle as the first parameter, so considered ze as the prefix
It would be better to have zet.
But would it be OK to have zet prefix and not have MetricHandles passed ?

Could you please provide your opinion on the above clarification.

Thanks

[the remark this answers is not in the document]
jandres742: zet is not about Metrics Handles but about whether the API belongs to Core or Tools. In this case, this API is expected to be used only for Tools, so zet would be better in my opinion.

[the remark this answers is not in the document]
joshuaranjan: Thanks @jandres742 
Renamed the API according to the module it belongs to.

jandres742 (commented):
+1

@wdamon-intel : please review.

jandres742:
@matcabral : please review.

scripts/tools/metricExportData.yml
```diff
@@ -0,0 +1,129 @@
...
+type: struct
+desc: "Metrics calculation descriptor"
+version: "1.6"
+class: $tMetric
+name: $t_metric_calculate_exp_desc_t
+base: $t_base_desc_t
+members:
+    - type: uint32_t
+      name: "rawReportSkipCount"
+      desc: "[in] number of reports to skip during calculation"
```
matcabral, line 49: minor: the description should be more explicit saying where are the reports skipped taken from. Beginning, end, middle?

joshuaranjan:
Hi @wdamon-intel @jandres742 
Would it be possible to get this in 1.6.x ?

wdamon-intel:
@joshuaranjan I've updated the project and milestone to v1.6.x.

jandres742 (approved):

Issue #84, opened 2023-02-24
Promote multiple metrics calculation API and deprecate the single value one 


matcabral:
Request is to promote zetMetricGroupCalculate**Multiple**MetricValuesExp() to no longer be an experimental extension and deprecate zetMetricGroupCalculateMetricValues(). The rationale behind this request is based on the inherent limitation in zetMetricGroupCalculateMetricValues() to calculate metrics values when the raw data collected contains counters from multiple sub-devices. Which today is not documented as an API  limitation.  A clear example is when collecting metrics with features like “implicit scaling” enabled in devices that have multiple sub-devices. To satisfy this gap zetMetricGroupCalculateMultipleMetricValuesExp() was added as an experimental extension around a year ago. Since then the API has not been modified. It is worth to mention that the multiple metrics set API is backward compatible, meaning that it can be used to collect metrics for devices with no sub-devices or at a sub-device level itself. In such cases pSetCount is simply set to 1.

joshuaranjan:
+1

jandres742:
PR in https://github.com/oneapi-src/level-zero-spec/pull/96

PR #96, opened 2023-03-09
Promote zetMetricGroupCalculateMultipleMetricValuesExp to standard

jandres742:
zetMetricGroupCalculateMultipleMetricValuesExp is kept and marked as deprecated to allow for a smooth transtion on applications. zetMetricGroupCalculateMultipleMetricValuesExp will be removed on next minor version.

zetMetricGroupCalculateMetricValues is also marked as deprecated, as zetMetricGroupCalculateMultipleMetricValues provides better functionality. zetMetricGroupCalculateMetricValues will be kept until next major revision to avoid breaking backward compatibility.

Resolves: #84

### DIFF
--- a/scripts/tools/PROG.rst
+++ b/scripts/tools/PROG.rst
@@ -589,23 +589,22 @@ Sample below shows how to calculate and process multiple metric values.
        // Example showing how to process the metric values
 
        // Setup
-       uint32_t metricCount = 0;
+       uint32_t metricCountInMetricGroup= 0;
        ${t}MetricGet(hMetricGroup, &metricCount, nullptr);
 
-       ${t}_metric_handle_t* phMetrics = malloc(metricCount * sizeof(${t}_metric_handle_t));
+       ${t}_metric_handle_t* phMetrics = malloc(metricCountInMetricGroup* sizeof(${t}_metric_handle_t));
        ${t}MetricGet(hMetricGroup, &metricCount, phMetrics);
 
        // Loop over the collected metrics
        uint32_t startIndex = 0;
        for (uint32_t setIndex = 0; setIndex < setCount; setIndex++) {
 
-           // Processing each metric data is the same as for the single
-           // calculate metric values
-           const uint32_t metricCountForDataIndex = metricCountPerSet[setIndex];
-           const uint32_t reportCount = metricCountForDataIndex / metricCount;
+           // Process each metric value for every report collected
+           const uint32_t metricsCountInSet= metricCountPerSet[setIndex];
+           const uint32_t reportCount = metricsCountInSet/ metricCount;
            for (uint32_t report = 0; report < reportCount; report++) {
-               for (uint32_t metric = 0; metric < metricCount ; metric++) {
-                   const size_t metricIndex = report * metricCount + metric;
+               for (uint32_t metric = 0; metric < metricCountInMetricGroup; metric++) {
+                   const size_t metricIndex = report * metricCountInMetricGroup+ metric;
                    process_metric_value(metricValues[startIndex + metricIndex]));
                }
            }
--- a/scripts/tools/PROG.rst
+++ b/scripts/tools/PROG.rst
@@ -576,12 +576,12 @@ Sample below shows how to calculate and process multiple metric values.
 
 .. parsed-literal::
 
-       // Get data count and total metric count
+       // Get metrics set count and total metrics count
        uint32_t setCount = 0;
        uint32_t metricValueCount = 0;
        ${t}MetricGroupCalculateMultipleMetricValues(hMetricGroup, type, rawDataSize, pRawData, &setCount, &metricValueCount, nullptr, nullptr);
 
-       // Get metric counts and metric values
+       // Get the metrics count per set and metrics values
        std::vector<uint32_t> metricCountPerSet(setCount);
        std::vector<${t}_typed_value_t> metricValues(metricValueCount);
        ${t}MetricGroupCalculateMultipleMetricValues(hMetricGroup, type, rawDataSize, pRawData, &setCount, &metricValueCount, metricCountPerSet.data(), metricValues.data());
@@ -595,13 +595,13 @@ Sample below shows how to calculate and process multiple metric values.
        ${t}_metric_handle_t* phMetrics = malloc(metricCount * sizeof(${t}_metric_handle_t));
        ${t}MetricGet(hMetricGroup, &metricCount, phMetrics);
 
-       // This loop over metric data is new for this extension
+       // Loop over the collected metrics
        uint32_t startIndex = 0;
-       for (uint32_t dataIndex = 0; dataIndex < setCount; dataIndex++) {
+       for (uint32_t setIndex = 0; setIndex < setCount; setIndex++) {
 
            // Processing each metric data is the same as for the single
            // calculate metric values
-           const uint32_t metricCountForDataIndex = metricCountPerSet[dataIndex];
+           const uint32_t metricCountForDataIndex = metricCountPerSet[setIndex];
            const uint32_t reportCount = metricCountForDataIndex / metricCount;
            for (uint32_t report = 0; report < reportCount; report++) {
                for (uint32_t metric = 0; metric < metricCount ; metric++) {
--- a/scripts/tools/PROG.rst
+++ b/scripts/tools/PROG.rst
@@ -568,6 +568,53 @@ The following pseudo-code demonstrates a basic sequence for metric calculation a
            free(phMetrics);
        }
 
+Calculating Multiple Metrics
+-----------
+
+${t}MetricGroupCalculateMultipleMetricValues can be used to calculate one or more sets of metric values from raw data.
+Sample below shows how to calculate and process multiple metric values.
+
+.. parsed-literal::
+
+       // Get data count and total metric count
+       uint32_t setCount = 0;
+       uint32_t metricValueCount = 0;
+       ${t}MetricGroupCalculateMultipleMetricValues(hMetricGroup, type, rawDataSize, pRawData, &setCount, &metricValueCount, nullptr, nullptr);
+
+       // Get metric counts and metric values
+       std::vector<uint32_t> metricCountPerSet(setCount);
+       std::vector<${t}_typed_value_t> metricValues(metricValueCount);
+       ${t}MetricGroupCalculateMultipleMetricValues(hMetricGroup, type, rawDataSize, pRawData, &setCount, &metricValueCount, metricCountPerSet.data(), metricValues.data());
+
+       // Example showing how to process the metric values
+
+       // Setup
+       uint32_t metricCount = 0;
+       ${t}MetricGet(hMetricGroup, &metricCount, nullptr);
+
+       ${t}_metric_handle_t* phMetrics = malloc(metricCount * sizeof(${t}_metric_handle_t));
+       ${t}MetricGet(hMetricGroup, &metricCount, phMetrics);
+
+       // This loop over metric data is new for this extension
+       uint32_t startIndex = 0;
+       for (uint32_t dataIndex = 0; dataIndex < setCount; dataIndex++) {
+
+           // Processing each metric data is the same as for the single
+           // calculate metric values
+           const uint32_t metricCountForDataIndex = metricCountPerSet[dataIndex];
+           const uint32_t reportCount = metricCountForDataIndex / metricCount;
+           for (uint32_t report = 0; report < reportCount; report++) {
+               for (uint32_t metric = 0; metric < metricCount ; metric++) {
+                   const size_t metricIndex = report * metricCount + metric;
+                   process_metric_value(metricValues[startIndex + metricIndex]));
+               }
+           }
+
+           startIndex += metricCountForDataIndex;
+       }
+       assert(startIndex == metricValueCount);
+
+
 
 Program Instrumentation
 =======================
--- a/scripts/tools/metric.yml
+++ b/scripts/tools/metric.yml
@@ -165,6 +165,7 @@ class: $tMetricGroup
 name: CalculateMultipleMetricValues
 decl: static
 details:
+    - "Application must call firstly pSetCount and pMetricValueCount to know the size to allocate for pMetricCountPerSet and pMetricValues, respectively."
     - "This function calculates more than one set of metric values from a single data buffer.  There may be one set of metric values for each sub-device, for example."
     - "Each set of metric values may consist of a different number of metric values, returned as the metric value count."
     - "All metric values are calculated into a single buffer; use the metric counts to determine which metric values belong to which set."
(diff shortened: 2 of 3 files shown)


matcabral:
@dmichaelo1 @joshuaranjan

matcabral (approved):

scripts/tools/metric.yml
```diff
@@ -158,6 +159,53 @@ params:
...
+            if *pSetCount is greater than the number available in the raw data buffer, then the driver shall update the value with the actual number of metric values to be calculated.
+    - type: uint32_t*
+      name: pMetricCounts
+      desc: |
+            [in,out][optional][range(0, *pSetCount)] buffer of metric counts per metric set.
+    - type: "$t_typed_value_t*"
+      name: pMetricValues
+      desc: |
+            [in,out][optional][range(0, *pTotalMetricValueCount)] buffer of calculated metrics.
+            if *pSetCount is less than the number available in the raw data buffer, then driver shall only calculate that number of metric values.
```
bmyates, line 206: should pSetCount be pTotalMetricValueCount

scripts/tools/metric.yml
```diff
@@ -158,6 +159,53 @@ params:
...
+            if *pSetCount is greater than the number available in the raw data buffer, then the driver shall update the value with the actual number of metric sets to be calculated.
+            if *pSetCount is is less than the number available in the raw data buffer, then the driver shall calculate the metrics for up to that number of sub-devices.
+    - type: uint32_t*
+      name: pTotalMetricValueCount
+      desc: |
+            [in,out] pointer to number of the total number of metric values calculated, for all metric sets.
+            if *pSetCount is zero, then the driver shall update the value with the total number of metric values to be calculated.
+            if *pSetCount is greater than the number available in the raw data buffer, then the driver shall update the value with the actual number of metric values to be calculated.
+    - type: uint32_t*
+      name: pMetricCounts
```
bmyates, line 199: imo naming this something like `pMetricCountBuffer` may help distinguish this is a buffer and not a count like the previous two fields.

scripts/tools/multiMetricValues.yml
```diff
@@ -32,6 +32,7 @@ class: $tMetricGroup
 name: CalculateMultipleMetricValuesExp
 decl: static
 details:
+    - "1.6": "Note: This extension is deprecated and has now been promoted to $tMetricGroupCalculateMultipleMetricValues. It is expected to be removed on next minor version update."
```
bmyates, line 35: caution: removing experimental extensions may cause compatibility issues in DDI tables.

scripts/tools/metric.yml
matcabral: the point is that none of the APis in the spec for which there is a `count` parameter that returns the number of entries in the following buffer has `buffer` as part of the buffer name

scripts/tools/metric.yml
matcabral: good catch

joshuaranjan (commented):
Should we modifying [EXT_Exp_MultiMetricValues.rst](https://github.com/oneapi-src/level-zero-spec/blob/master/scripts/tools/EXT_Exp_MultiMetricValues.rst) in this PR ?

scripts/tools/metric.yml
jandres742: @matcabral this is the same as the experimental extension. @matcabral please confirm then what it should go here.

scripts/tools/multiMetricValues.yml
jandres742: @bmyates we are not removing it here but it will be eventually. This is an experimental extension, so there are no guarantees on backwards compatibility.

jandres742:
> Should we be modifying [EXT_Exp_MultiMetricValues.rst](https://github.com/oneapi-src/level-zero-spec/blob/master/scripts/tools/EXT_Exp_MultiMetricValues.rst) in this PR ?

that's a good suggestion. I will change it.

scripts/tools/multiMetricValues.yml
bmyates: Yes, understood.  However removing it will cause issues with DDI table and break compatibility of other experimental APIs. There will need to be some workarounds put into code generation to avoid this.   It may be easier to not remove it. 

https://github.com/intel-innersource/libraries.compute.oneapi.level-zero.loader/blob/21acb8b405a839e7ff09bd2a49472b217ba871a8/include/zet_ddi.h#L361

scripts/tools/metric.yml
joshuaranjan: @matcabral 
Implementation updates "pMetricValues"  for only for the "psetCount" provided.

jandres742: @matcabral , @joshuaranjan : could you suggest how the parameters should be named and described? in this interface, we have two count variables.

pSetCount is associated with pMetricCounts, could we suggest a different name for pMetricCounts to not have the "Count" in it?

pTotalMetricValueCount is associated with pMetricValues. So pTotalMetricValueCount  could be just called pMetricValueCount?

matcabral: @jandres742 In respect to the names, I think the current ones are accurate.  In terms of the description, I have some comments below:

>pSetCount is associated with pMetricCounts, could we suggest a different name for pMetricCounts to not have the "Count" in it?

An actual clearer name would be `pMetricCountPerSet` : Note that this is an array of sizes/counts, with each element being the number of metrics available in each `set` (defied by `pSetCount`). Each set can have different numbers. 

> pTotalMetricValueCount is associated with pMetricValues. So pTotalMetricValueCount could be just called pMetricValueCount?

Strictly speaking, your statement is correct. However, since there are few counts in the API the intention was to be upfront clear. I don’t have a strong opinion about either name. Maybe improving a more explicit description would allow changing :

` pointer to number of the total number of metric values calculated, it is the sum of all elements in buffer  pMetricCountPerSet . if count is zero...`

It is worth to note that in the original version of the API (zetMetricGroupCalculateMetricValues()) the parameter is called pMetricValueCount, so keeping the same name could be seen favorable. 


In terms of the description and behavior of description and behavior :

I saw that the pSetCount description missed a text describing it can be used to limit the number of sets to be calculated. So the following text should be appended at the end:

` if count is less than the number of available sets in the raw data buffer, then driver shall only calculate that number of sets. ` 

Still need to define how the user can limit the number of metrics to be calculated, let me drive that discussion by email and write the resolution here.

joshuaranjan: +1 for "pMetricCountPerSet" as suggested above

>if count is less than the number of available sets in the raw data buffer, then driver shall only calculate that number of sets.
@matcabral , In explanation about "count" in other instances, the spec covers 2 conditions
1. what happens when it is 0
2. what happens when count is greater than available.

Since both of these are covered for "pSetCount ", please share if we need the additional point that is suggested ?
Thanks

matcabral: > Since both of these are covered for "pSetCount ", please share if we need the additional point that is suggested ?

@joshuaranjan   yes:   

1.  0 is a query
2. greater than available, scopes to whats available 
3.  smaller than avaialble, limits to that number

scripts/tools/PROG.rst
```diff
@@ -568,6 +568,53 @@ The following pseudo-code demonstrates a basic sequence for metric calculation a
...
 
+Calculating Multiple Metrics
+-----------
+
+${t}MetricGroupCalculateMultipleMetricValues can be used to calculate one or more sets of metric values from raw data.
+Sample below shows how to calculate and process multiple metric values.
+
+.. parsed-literal::
+
+       // Get data count and total metric count
```
matcabral, line 579: `get metrics set count and total metrics count`

scripts/tools/PROG.rst
```diff
@@ -568,6 +568,53 @@ The following pseudo-code demonstrates a basic sequence for metric calculation a
...
+       uint32_t setCount = 0;
+       uint32_t metricValueCount = 0;
+       ${t}MetricGroupCalculateMultipleMetricValues(hMetricGroup, type, rawDataSize, pRawData, &setCount, &metricValueCount, nullptr, nullptr);
+
+       // Get metric counts and metric values
```
matcabral, line 584: `Get the metrics count per set and metrics values`

scripts/tools/PROG.rst
```diff
@@ -568,6 +568,53 @@ The following pseudo-code demonstrates a basic sequence for metric calculation a
...
+       // Example showing how to process the metric values
+
+       // Setup
+       uint32_t metricCount = 0;
+       ${t}MetricGet(hMetricGroup, &metricCount, nullptr);
+
+       ${t}_metric_handle_t* phMetrics = malloc(metricCount * sizeof(${t}_metric_handle_t));
+       ${t}MetricGet(hMetricGroup, &metricCount, phMetrics);
+
+       // This loop over metric data is new for this extension
```
matcabral, line 598: `Loop over the collected metrics`

scripts/tools/PROG.rst
```diff
@@ -568,6 +568,53 @@ The following pseudo-code demonstrates a basic sequence for metric calculation a
...
+       uint32_t startIndex = 0;
+       for (uint32_t dataIndex = 0; dataIndex < setCount; dataIndex++) {
```
matcabral, line 600: minor: it would be clearer to rename `dataIndex` as `setIndex`

scripts/tools/PROG.rst
```diff
@@ -568,6 +568,53 @@ The following pseudo-code demonstrates a basic sequence for metric calculation a
...
+
+           // Processing each metric data is the same as for the single
```
matcabral, line 602: Since we are deprecating the single set metrics calcuate, I think would be better to avoid references to it.  Better somthing like: 
`Process each metric value for every report collected`

scripts/tools/PROG.rst
```diff
@@ -568,6 +568,53 @@ The following pseudo-code demonstrates a basic sequence for metric calculation a
...
+           // calculate metric values
+           const uint32_t metricCountForDataIndex = metricCountPerSet[dataIndex];
```
matcabral, line 604: `metricCountForDataIndex` rename to `metricsCountInSet`

scripts/tools/PROG.rst
```diff
@@ -568,6 +568,53 @@ The following pseudo-code demonstrates a basic sequence for metric calculation a
...
+
+       // Get metric counts and metric values
+       std::vector<uint32_t> metricCountPerSet(setCount);
+       std::vector<${t}_typed_value_t> metricValues(metricValueCount);
+       ${t}MetricGroupCalculateMultipleMetricValues(hMetricGroup, type, rawDataSize, pRawData, &setCount, &metricValueCount, metricCountPerSet.data(), metricValues.data());
+
+       // Example showing how to process the metric values
+
+       // Setup
+       uint32_t metricCount = 0;
```
matcabral, line 592: `metricCount`  -> `metricCountInMetricGroup` 

Having so many counts in flight, having more explicit names would help.

scripts/tools/metric.yml
```diff
@@ -165,6 +165,7 @@ class: $tMetricGroup
 name: CalculateMultipleMetricValues
 decl: static
 details:
+    - "Application must call firstly pSetCount and pMetricValueCount to know the size to allocate for pMetricCountPerSet and pMetricValues, respectively."
```
matcabral, line 168: `"Application must query values of pSetCount and pMetricValueCount to know the size to allocate for pMetricCountPerSet and pMetricValues, respectively."`

scripts/tools/metric.yml
```diff
@@ -185,25 +186,29 @@ params:
...
+            if *pMetricValueCount is zero, then the driver shall update the value with the total number of metric values to be calculated.
+            if *pMetricValueCount is greater than the number available in the raw data buffer, then the driver shall update the value with the actual number of metric values to be calculated.
     - type: uint32_t*
-      name: pMetricCounts
+      name: pMetricCountPerSet
       desc: |
             [in,out][optional][range(0, *pSetCount)] buffer of metric counts per metric set.
+            Driver shall update the value with the number of metric values calculated for that set.
+            If count for a given set is greater than the number of metric values available in the raw data buffer in that set, then the driver
+            shall update the value with the actual number of metric values to be calculated for that set.
```
matcabral, line 205: `then the driver shall update the value with the actual number of metric values calculated for that set.` 

remove "to be"

scripts/tools/PROG.rst
```diff
@@ -568,6 +568,53 @@ The following pseudo-code demonstrates a basic sequence for metric calculation a
...
+
+       // Get data count and total metric count
+       uint32_t setCount = 0;
+       uint32_t metricValueCount = 0;
+       ${t}MetricGroupCalculateMultipleMetricValues(hMetricGroup, type, rawDataSize, pRawData, &setCount, &metricValueCount, nullptr, nullptr);
+
+       // Get metric counts and metric values
+       std::vector<uint32_t> metricCountPerSet(setCount);
+       std::vector<${t}_typed_value_t> metricValues(metricValueCount);
+       ${t}MetricGroupCalculateMultipleMetricValues(hMetricGroup, type, rawDataSize, pRawData, &setCount, &metricValueCount, metricCountPerSet.data(), metricValues.data());
```
joshuaranjan, line 587: Can we add a line where we are modifying the value for metricCountPerSet.
@matcabral 
Should the below be OK ?
metricCountPerSet[0] = metricValueCount / 2;
metricCountPerSet[1] = metricValueCount / 2;

matcabral: I don't expect the use case of limiting metrics per set to be very common. So, IMHO adding it in the example will unnecessarily complicate the common use case.

scripts/tools/metric.yml
```diff
@@ -158,6 +159,58 @@ params:
...
+            if *pSetCount is greater than the number available in the raw data buffer, then the driver shall update the value with the actual number of metric sets to be calculated.
+            if *pSetCount is less than the number of sets available, then driver shall only calculate that number of sets.
+    - type: uint32_t*
+      name: pMetricValueCount
+      desc: |
+            [in,out][optional] pointer to the total number of metric values calculated, it is the sum of all elements in pMetricCountPerSet array.
+            if *pMetricValueCount is zero, then the driver shall update the value with the total number of metric values to be calculated.
+            if *pMetricValueCount is greater than the number available in the raw data buffer, then the driver shall update the value with the actual number of metric values to be calculated.
+    - type: uint32_t*
+      name: pMetricCountPerSet
```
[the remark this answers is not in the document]
matcabral: complete text: 
`[in,out][optional][range(0, *pSetCount)] buffer of metric counts per metric set.
            Driver shall update the value with the number of metric values calculated for that set. If count for a given set is greater than the number of metric values available in the raw data buffer in that set, then the driver shall update the value with the actual number of metric values calculated for that set. If count for a given set is less than the number of metric values available in the raw data buffer in that set (count must be multiple of the number of metrics in the compute group), then driver shall only calculate that number of metric values for that set.
`

joshuaranjan:
lgtm

jandres742:
@matcabral @joshuaranjan : addressed comments. Please check.

scripts/tools/PROG.rst
```diff
@@ -589,23 +589,22 @@ Sample below shows how to calculate and process multiple metric values.
        // Example showing how to process the metric values
 
        // Setup
-       uint32_t metricCount = 0;
+       uint32_t metricCountInMetricGroup= 0;
        ${t}MetricGet(hMetricGroup, &metricCount, nullptr);
```
joshuaranjan, line 593: [Minor] metricCount to metricCountInMetricGroup

scripts/tools/PROG.rst
```diff
@@ -589,23 +589,22 @@ Sample below shows how to calculate and process multiple metric values.
...
        uint32_t startIndex = 0;
        for (uint32_t setIndex = 0; setIndex < setCount; setIndex++) {
 
-           // Processing each metric data is the same as for the single
-           // calculate metric values
-           const uint32_t metricCountForDataIndex = metricCountPerSet[setIndex];
-           const uint32_t reportCount = metricCountForDataIndex / metricCount;
+           // Process each metric value for every report collected
+           const uint32_t metricsCountInSet= metricCountPerSet[setIndex];
+           const uint32_t reportCount = metricsCountInSet/ metricCount;
```
joshuaranjan, line 604: Here as well metricCount to metricCountInMetricGroup

scripts/tools/PROG.rst
```diff
@@ -589,23 +589,22 @@ Sample below shows how to calculate and process multiple metric values.
        // Example showing how to process the metric values
 
        // Setup
-       uint32_t metricCount = 0;
+       uint32_t metricCountInMetricGroup= 0;
        ${t}MetricGet(hMetricGroup, &metricCount, nullptr);
 
-       ${t}_metric_handle_t* phMetrics = malloc(metricCount * sizeof(${t}_metric_handle_t));
+       ${t}_metric_handle_t* phMetrics = malloc(metricCountInMetricGroup* sizeof(${t}_metric_handle_t));
        ${t}MetricGet(hMetricGroup, &metricCount, phMetrics);
```
joshuaranjan, line 596: Please modify here as well

joshuaranjan (commented):
minor points

matcabral (approved):

matcabral (changes requested):
Reverting my approval since @joshuaranjan suggested an approach that would provide much more flexibility at the time of calculating metrics. Having a descriptor struct as an input:

``
zet_metric_calculate_ desc_t {
                zet_structure_type_t stype; 
                const void *pNext ;
                uint32_t rawReportSkipCount; // number of reports to skip during calculation
	…
	…
}
`` 

Therefore, we need to review offline what this structure should look like.

wdamon-intel:
Too much ongoing discussion for this one to make v1.6. Moving to v1.7.

wdamon-intel:
@matcabral @joshuaranjan any additional questions/concerns?

matcabral:
Hi @wdamon-intel , we have pending the analysis for supporting a new parameters of a description structure and the components in the structure itself. It is worth the review it in detail before promoting the API.

matcabral:
@wdamon-intel please close this PR, this API will be superseded with a more flexible approach.

wdamon-intel:
Closing as Won't Implement, per discussion.

PR #100, opened 2023-03-20
Move restrictions about image buffers to under ze_image_type_t

lyu:
Media formats are defined using the layout member, not the format type member

### DIFF
--- a/scripts/core/image.yml
+++ b/scripts/core/image.yml
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2019-2022 Intel Corporation
+# Copyright (C) 2019-2023 Intel Corporation
 #
 # SPDX-License-Identifier: MIT
 #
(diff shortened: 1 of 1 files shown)


scripts/core/image.yml
```diff
@@ -173,10 +173,10 @@ name: $x_image_format_t
 members:
     - type: $x_image_format_layout_t
       name: layout
-      desc: "[in] image format component layout"
+      desc: "[in] image format component layout. Media formats can't be used for $X_IMAGE_TYPE_BUFFER."
```
jandres742, line 176: @wdamon-intel : don't know if we should have this guarded with a version or not.

wdamon-intel: - What issue is this addressing?
- Why is this a restriction?
- Could media format layouts ever be supported for buffer types?

Instead of wording this as something that *can't* be done, it should be rephrased to something that *should* or *must* be done. For example, in the description of `layout`, state that "media formats are a type of layout", and under type `type` description, state that "media format layouts are unsupported for $X_IMAGE_TYPE_BUFFER", and (possibly?) wrap it with a version guard...

wdamon-intel (approved):
LGTM

jandres742 (approved):

Issue #57, opened 2023-01-08
Global Contexts in L0


jandres742:
L0 objects should be created with a specific context, and with that we maintain isolation of resources.

However, main issue on multi-library scenarios, is that we have objects being passed between libraries (like USM allocations), from one library to another, but the context is not known.

One possible solution is to return the context so a library can create a queue or list based on that context:

```cpp
// MPI receives USM pointer from another library
zeMemGetAllocProperties(..., &zeContext);
// now use the context to create the queue and list
zeCommandQueueCreate(zeContext, ...)
zeCommandListCreate(zeContext, ...)
...
```

however, with this, libraries are forced to have lazy initialization of resources. To allow having eager initialization of resources, something like an **L0 Global Context** is needed, so libraries can use USM pointers with L0 objects created with any other context, like queues and lists.

So, should we define an L0 global context? we could have something like

```cpp
zeDriverGetGlobalContext(hDriver, &hContext);
```
and when using that context, L0 doesnt enforce separation of resources, allowing USM pointers to be used with lists, queues, and other USM allocations created with any context,

CC: @zackwaters , @wdamon-intel , @bashbaug .

cbelusar:
We (communication libraries team) have concerns about this requirement. Can you involve us in design discussions going forward involving this proposal? 
Please include ( Alex Brooks - @abrooks98, Gengbin Zheng - @zhenggb72, Anatoliy Rozanov - @irozanova, Dmitry Durnov - @ddurnov )

wdamon-intel:
Instead of global context, one proposal is to make it possible to associate one or more contexts with a shared context group, via a new interface:

```c
// Define a shared context group of existing contexts
//
// hDriver, nonnull, handle to the driver object
// pCount, nonnull, number of contexts in the array of context handles
// phContexts, nonnull, array of context handles to associate within a context share-group
// phContextShared, nullable, pointer to the handle of the associated context share-group found or created by the runtime
ze_result_t zeContextSharedGet(ze_driver_handle_t hDriver, uint32_t *pCount, ze_context_handle_t *phContexts, ze_context_shared_handle_t *phContextShared); 

// Add a context to an existing shared context group
ze_result_t zeContextSharedAddContext(ze_context_shared_handle_t hContextShared, ze_context_handle_t hContext);

// Get the share-group of a context.
// Returns a null pointer in phContextShared if the context is not part of a share-group.
[...]

PR #101, opened 2023-03-21
Add experimental extensions for shared-contexts

jandres742:
Resolves: #57

### DIFF
--- a/scripts/core/EXT_Exp_SharedContexts.rst
+++ b/scripts/core/EXT_Exp_SharedContexts.rst
@@ -1,0 +1,51 @@
+<%
+import re
+from templates import helper as th
+%><%
+    OneApi=tags['$OneApi']
+    x=tags['$x']
+    X=x.upper()
+%>
+:orphan:
+
+.. _ZE_experimental_shared_contexts:
+
+=====================================
+ Shared-Contexts Extension
+=====================================
+
+API
+----
+
+* Enumerations
+
+
+    * ${x}_shared_contexts_exp_version_t
+
+
+* Functions
+
+    * ${x}MemGetAllocPropertiesExp
+    * ${x}DriverGetSharedContextExp
+    * ${x}DriverAddContextToSharedExp
+
+Shared Contexts
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Contexts provide a way to isolate resources in an Level Zero application. Specifically,
+USM allocations are created with a context handle and can be used only with
+other objects (like command lists) that have been created with the same context.
+
+To be able to use those resources in applications where USM pointers
+(and not their contexts) are exchanged between multiple libraries, the concept
+of shared contexts interfaces is defined.
+
+A shared context is an Level Zero context handle that encompases a set of contexts
+handles. Any Level Zero object belonging to one of the contexts
+associated with the shared context can be used and interact with an object of the other
+peer contexts. Ownership of L0 objects is not transferred, however.
+
+A given context can only belong to a single shared context, and once created,
+a shared context can be used normally in any API. Objects created using
+the shared context handle implicitly belong to all the contexts associated with
+it. A shared context remains valid until its last context is destroyed.
--- a/scripts/core/sharedContext.yml
+++ b/scripts/core/sharedContext.yml
@@ -1,0 +1,88 @@
+#
+# Copyright (C) 2023 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+#
+# See YaML.md for syntax definition
+#
+--- #--------------------------------------------------------------------------
+type: header
+desc: "Intel $OneApi Level-Zero Extension for shared contexts."
+version: "1.6"
+--- #--------------------------------------------------------------------------
+type: macro
+desc: "Shared Contexts Extension Name"
+version: "1.6"
+name: $X_SHARED_CONTEXTS_EXP_NAME
+value: '"$X_experimental_shared_contexts"'
+--- #--------------------------------------------------------------------------
+type: enum
+desc: "Shared Contexts Extension Version(s)"
+version: "1.6"
+name: $x_shared_contexts_exp_version_t
+etors:
+    - name: "1_0"
+      value: "$X_MAKE_VERSION( 1, 0 )"
+      desc: "version 1.0"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Defines a shared context for a set of context handles."
+version: "1.6"
+class: $xDriver
+name: GetSharedContextExp
+details:
+    -  |
+       If a shared context has been previously created for the contexts passed in the array,
+       and the shared-context is the same for all, then that one is returned.
+    -  |
+       If none of the contexts in the array belong to a shared context, then a shared context is created
+       encompassing all the contexts in the array.
+    -  |
+       If some of the contexts in the array are not yet associated with any shared context, and the rest
+       belong to a single shared context, then all are added to the same shared context.
+    -  |
+       If some of the contexts in the array already belong to different shared contexts,
+       then nullptr is returned.
+    -  |
+       If only one context is passed, then its shared context is returned if it already exists.
+       If not, then a new shared context is created and returned. Additional contexts
+       can then later be added to this one using $xDriverAddContextToSharedExp.
+    - "If only one context is passed, and that context is a shared context, then the same is returned."
+    - "The application may call this function from simultaneous threads with the same context handle."
+    - "The implementation of this function should be lock-free."
+params:
+    - type: $x_driver_handle_t
+      name: hDriver
+      desc: "[in] handle of driver object"
+    - type: uint32_t
+      name: count
+      desc: "[in] number of context passed in hContexts array"
+    - type: "$x_context_handle_t*"
+      name: phContexts
+      desc: "[in][range(0, count)] array of contexts to create the new shared-context with"
+    - type: "$x_context_handle_t*"
+      name: phSharedContext
+      desc: "[out] pointer to the shared context object"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Adds a context to an existing shared context."
+version: "1.6"
+class: $xDriver
+name: AddContextToSharedExp
+details:
+    -  |
+       The context must not belong already to another shared context. If so, $X_RESULT_ERROR_INVALID_ARGUMENT
+       is returned.
+    - "The application may call this function from simultaneous threads with the same context handle."
+    - "The implementation of this function should be lock-free."
+params:
+    - type: $x_driver_handle_t
+      name: hDriver
+      desc: "[in] handle of driver object"
+    - type: "$x_context_handle_t"
+      name: hSharedContext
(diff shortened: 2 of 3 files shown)


jandres742:
@wdamon-intel : please check.

scripts/core/sharedContext.yml
```diff
@@ -0,0 +1,88 @@
...
+      desc: "[in][range(0, count)] array of contexts to create the new shared-context with"
+    - type: "$x_context_handle_t*"
+      name: phSharedContext
+      desc: "[out] pointer to the shared context object"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Adds a context to an existing shared context."
+version: "1.6"
+class: $xDriver
+name: AddContextToSharedExp
```
MichalMrozek, line 71: how is context removed from shared context?

wdamon-intel: Does the API need a mechanism for this, or can the driver/runtime do it during context teardown?

jandres742: @MichalMrozek thanks. IWe could add zeDriverRemoveContextFromShared.

scripts/core/EXT_Exp_SharedContexts.rst
```diff
@@ -0,0 +1,51 @@
...
+(and not their contexts) are exchanged between multiple libraries, the concept
+of shared contexts interfaces is defined.
+
+A shared context is an Level Zero context handle that encompases a set of contexts
+handles. Any Level Zero object belonging to one of the contexts
+associated with the shared context can be used and interact with an object of the other
+peer contexts. Ownership of L0 objects is not transferred, however.
+
+A given context can only belong to a single shared context, and once created,
+a shared context can be used normally in any API. Objects created using
```
MichalMrozek, line 49: I see the cluase that objects created using shared context belong to all context.

What if context is added later to shared context, are those resource allocated prior to this addition are now available in this newly added context?

scripts/core/sharedContext.yml
```diff
@@ -0,0 +1,88 @@
...
+       If a shared context has been previously created for the contexts passed in the array,
+       and the shared-context is the same for all, then that one is returned.
+    -  |
+       If none of the contexts in the array belong to a shared context, then a shared context is created
+       encompassing all the contexts in the array.
+    -  |
+       If some of the contexts in the array are not yet associated with any shared context, and the rest
+       belong to a single shared context, then all are added to the same shared context.
+    -  |
+       If some of the contexts in the array already belong to different shared contexts,
```
MichalMrozek, line 44: This sentence contradicts with previous sentence.

Let's say there is shared context 1 with context 1
Now we try to create shared context , with context 1 and context 2, previous sentence 

"
If some of the contexts in the array are not yet associated with any shared context, and the rest
       belong to a single shared context, then all are added to the same shared context."

Tells us that context 2 will be added to shared context 1.

Now this sentence tells us that if context 1 belongs to shared context 1, then nullptr is returned.

scripts/core/sharedContext.yml
```diff
@@ -0,0 +1,88 @@
...
+       then nullptr is returned.
+    -  |
+       If only one context is passed, then its shared context is returned if it already exists.
+       If not, then a new shared context is created and returned. Additional contexts
+       can then later be added to this one using $xDriverAddContextToSharedExp.
+    - "If only one context is passed, and that context is a shared context, then the same is returned."
```
MichalMrozek, line 50: Does this API allow passing shared context as input ?

scripts/core/EXT_Exp_SharedContexts.rst
```diff
@@ -0,0 +1,51 @@
...
+* Functions
+
+    * ${x}MemGetAllocPropertiesExp
+    * ${x}DriverGetSharedContextExp
+    * ${x}DriverAddContextToSharedExp
+
+Shared Contexts
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Contexts provide a way to isolate resources in an Level Zero application. Specifically,
```
MichalMrozek, line 35: what's the use case here?
We need to be careful to make sure that we do not change the basic paradigm, where resources allocated in one context will be automatically available in another context.

If apps want this kind of behaviour then it needs to allocate shared context from multiple contexts and use this shared context for resource / program /kernel allocation.

scripts/core/sharedContext.yml
MichalMrozek: We need to be careful with this extension to not relax the current boundaries too much. One problem that I see is that it allows resources to be avaiblle automatically outside of the context where they were allocated.

jandres742: @MichalMrozek . Correct, that's the main idea. What happens is that we could have multi-library applications (MPI + SYCL + MKL), where USM allocations are created by one library and then their pointers exchanged with others (like SYCL allocating the memory, and the MPI using it in accelerator collectives). So the only way they could use those pointers coming from other library is having a way to share the resource with their own context, which is where these APIs come into play, allowing users to explicitly request such sharing.

This is needed to start enforcing per-context resource isolation in the L0 GPU driver. We need to either add this or abandon for completely the idea of isolating resources per context.

scripts/core/EXT_Exp_SharedContexts.rst
jandres742: @MichalMrozek . Correct. If user already created a shared-context, it means they have explicitly opted-in for resource sharing, so it is naturally to assume that any resource created with the shared-context can be used with any of the contexts in the set.

Do you think there could be problems with that? if so, then we could just add a prohibition that no resources cannot be created using the shared-context.

scripts/core/sharedContext.yml
jandres742: @MichalMrozek 
in your example, context 1 belongs to shared-context 1. That is this sentence:
> If only one context is passed, then its shared context is returned if it already exists.
       If not, then a new shared context is created and returned

Now, context 2 is added to shared context 1, which is this sentence:
>  If some of the contexts in the array are not yet associated with any shared context, and the rest
       belong to a single shared context, then all are added to the same shared context.

Now, the sentence you are referring to says:
> If some of the contexts in the array already belong to different shared contexts,

doesn't apply here, as context 1 and context 2 belong to the same shared-context. This sentence would apply if there's a context 3 belonging to another shared-context, and then the following call is made

GetSharedContextExp({context1, context2, context3}) 

in this case, then as the sentence says, a null shared pointer would be returned, as all them already belong to different shared contexts.

scripts/core/sharedContext.yml
jandres742: @MichalMrozek : I dont see why not. In that case, the same shared-context would be returned. However, I see there could be a confusion if the interface is called with a mix of shared contexts and regular contexts, in that case, what the behavior should be? we could disallow that case but indicating that if the API is called with 1 or more shared-contexts, then nullptr is returned, which means, we dont create a shared-context of multiple shared-contexts.

scripts/core/sharedContext.yml
jandres742: @MichalMrozek  maybe we could clarify saying something like:

"If all contexts in the array already belong to shared-contexts, and those shared-context are not the same, then nullptr is returned"

does that sound more accurate?

scripts/core/EXT_Exp_SharedContexts.rst
jandres742: @MichalMrozek normal contexts would be the typical case, but I dont see why a shared context couldn't be used, but again, if we think there could be problems, we could add the restriction.

scripts/core/sharedContext.yml
MichalMrozek: Context is pretty fundamental approach in the API that provides encapsulation.
Enabling dynamic sharing is very challenging to deliver, basically all resources that you allocated prior to creation of shared context, now need to be re-evaluated on every subsequent submit, which would substantially add to overheads.

There are also other ways to solve this:

1. use import / export

With export / import notion you can still export resource from one context and import it to another.
It also clearly defines lifecycle and it connected to the resource.

2. add default context

Another way to solve this is to allow default context per driver instance. This would be simple getDefaultContext API that returns default context, SYCL uses it for all allocations.

scripts/core/EXT_Exp_SharedContexts.rst
jandres742: @MichalMrozek the idea is as you say above "where resources allocated in one context will be automatically available in another context.". The basic idea is to allow users to share resources (previously created or new) among contexts.

Once a shared-context has been created, then any resource previously or later created with one of those contexts is automatically available in the other contexts belonging to the same shared-context.

Whether the new resources are created with the shared context or with one of the contexts ultimately doesnt make a difference, since at the end they are shared among of all them.

Now, if we do as you mention above, that the only resources to be shared are those created with the shared-context, that could cause a problem of data races. Imagine you have two threads, one creating resources with context 0 and other with context 1. At some point, thread 1 creates a shared-context with the two contexts1.

If we do as you say, then only the resources created previously to thread 1 creating the shared context would be shared, but after that, thread 0 would have to use the shared context, but how thread 0 would know when or if a shared-context was created? I think that is overly complicated.

By creating the share-context, the application has already told the driver that wants this kind of behavior and that resources belonging to those contexts would be shared.

scripts/core/sharedContext.yml
jandres742: thanks @MichalMrozek . 

The idea of the default context is good, but I think that @wdamon-intel proposed this approach since it is more explicit. With default context, you have a default context belonging to all contexts, but with these APIs, user is doing a more explicitly opt-in for this feature. @wdamon-intel can elaborate further.

I like the idea of the import/export. Basically, instead of creating a shared-context and having all resources being shared, with import/export the sharing is more fine granular and specific to a given resources, instead of all. Also, we avoid the issues with shared-contexts that we are discussing here.

We could have the following:

```cpp
zeContextExportResource(hContext, *void handle)
the USM pointer or handle passed as void is marked by current library as exportable,
meaning that it can be imported by other contexts.

zeContextImportResource(hContext, *void handle)
the USM pointer or handle passed as void is imported into the context to be used with
any resource created with it.
```

@MichalMrozek is that more or less what you were thinking?

@wdamon-intel what do you think?

wdamon-intel: The basic idea was to create a mechanism by which libraries/instances (within the same process address space) could share resources while retaining isolation/separation as there is today. The "shared context" concept comes from OpenGL. There, a context is created with another context handle and they belong to the same "group" as far as the implementation is concerned. **There is no such thing as an explicit "shared context object"**. The implementation maintains the resource isolation between contexts, but contexts part of a share-group may directly access (most of) one-another's objects. The only way a context joins a group is at creation time, and the only way it leaves a group is by being destroyed.

To keep it simple, I think we could:

1. Abandon this proposal and just leverage import/export. Or,
2. Simplify this proposal and adopt some of the OpenGL share group rules discussed above.

wdamon-intel:
Closing this PR due to inactivity. The original issue will also be closed, and a new proposal will have to be drafted.

Issue #68, opened 2023-02-07
Define layout for ze_ipc_mem_handle_t?


jandres742:
ze_ipc_mem_handle_t is defined as an opaque data. however, drivers like L0 GPU driver, have a specific layout that applications might need to be aware of. For instance, applications might need to know:

- does ze_ipc_mem_handle_t contain only the dma-buf file descriptor in case of linux?
- can I zero the ze_ipc_mem_handle_t and copy the fd, or is there any data inside that needs to be kept that has been previously set by the driver?

So the question is: should we define a layout for ze_ipc_mem_handle_t? One example could be:

char data[ZE_MAX_IPC_HANDLE_SIZE][¶](https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=ipc#_CPPv4N19ze_ipc_mem_handle_t4dataE)

ZE_MAX_IPC_HANDLE_SIZE is 64B:

0 - 7: file descriptor
8: memory type
9 - 63: reserved

same could apply for ze_ipc_event_pool_handle_t, which drivers could use to encode other information:
0 - 7: file descriptor
8: num of events
9: flags
10:  reserved

However, another point of view could be: should we define this in the spec, or should we leave driver implementations to define that layout?

MichalMrozek:
why we cannot have it as opaque data?
Relying on specific driver implementation details is highly error prone and prevents any driver changes.

jandres742:
@MichalMrozek because applications may need that information.
For instance, for IPC memory: it is not opaque if L0 driver implementation is using a dma-buf handle. Applications need to know a dma-buf handle is used, need to extract it from the IPC handle, and pass it using specific Linux mechanisms, given dma-buf security restrictions.

So valid questions are: where the dma-buf handle is in the 64B of IPC handle's  data[ZE_MAX_IPC_HANDLE_SIZE], at the beginning? the end? is there any other information in that handle that the driver is passing between processes (L0 GPU driver is doing that). and of course, at the moment we say where the dma-buf fd is in the data array, it is not opaque anymore.

but again, maybe all this is very specific to L0 GPU driver implementation, and not needed in the spec, and if an L0 driver implementation needs applications to know how the IPC handle layout, then such documentation should be in the L0 driver implementation, not the spec.

MichalMrozek:
On Kernels 5.6 or greater Level Zero can do :

zeMemGetIpcHandle

call PrimeHandleToFd
set FD in ze_ipc_mem_handle_t
[...]

PR #105, opened 2023-03-27
Add helper functions for IPC handle

jandres742:
Resolves: #68

Define two new functions:

zeMemGetIpcHandleFromFileDescriptorExp
zeMemGetFileDescriptorFromIpcHandleExp

to allow users to construct an IPC memory handle from a file descriptor, or to get the file descriptor out of an IPC memory handle, respectively.

This allows applications to not need to know the details of how the driver creates the opaque IPC handle, and to intermix IPC opaque APIs (zeMemGetIpcHandle, zeMemOpenIpcHandle, etc) with explicit file descriptor APIs (e.g., ze_external_memory_export_fd_t).

### DIFF
--- a/scripts/core/memory.yml
+++ b/scripts/core/memory.yml
@@ -306,6 +306,49 @@ params:
       desc: "[out] Returned IPC memory handle"
 --- #--------------------------------------------------------------------------
 type: function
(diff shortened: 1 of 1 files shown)


wdamon-intel (approved):
LGTM

PR #113, opened 2023-04-11
Update metrics timer resolution to cycle/sec

matcabral:


matcabral:
@jandres742 @wdamon-intel

PR #83, opened 2023-02-24
Update value of ZES_MAX_RAS_ERROR_CATEGORY_COUNT macro

mraghuwa:

PR #35, opened 2022-11-02
RAS Category and Fabric API added for spec v1.5

jandres742:
Signed-off-by: Vaidyanathan Krishnamoorthy <vaidyanathan.krishnamoorthy@intel.com>

PR #117, opened 2023-04-25
Revert RAS Category and Fabric API commits (#35, #83)

wdamon-intel:
This reverts commits:
  - 7de96547c9b3bc190606eb0a3327e465e6c52459
  - 5ced682e98bf10002d2c3e519373a312f883dadd

...due to a binary incompatibility regression.

### DIFF
--- a/scripts/sysman/fabric.yml
+++ b/scripts/sysman/fabric.yml
@@ -185,26 +185,6 @@ members:
       name: "txCounter"
       desc: "[out] Monotonic counter for the number of bytes transmitted (sum of all lanes). This includes all protocol overhead, not only the GPU traffic."
 --- #--------------------------------------------------------------------------
-type: struct
-desc: "Fabric Port Error Counters"
-class: $sFabricPort
-name: $s_fabric_port_error_counters_t
-base: $s_base_properties_t
-version: "1.5"
-members:
-    - type: uint64_t
-      name: "linkFailureCount"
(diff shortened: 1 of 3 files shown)


scripts/sysman/fabric.yml
```diff
@@ -185,26 +185,6 @@ members:
       name: "txCounter"
       desc: "[out] Monotonic counter for the number of bytes transmitted (sum of all lanes). This includes all protocol overhead, not only the GPU traffic."
 --- #--------------------------------------------------------------------------
-type: struct
-desc: "Fabric Port Error Counters"
-class: $sFabricPort
-name: $s_fabric_port_error_counters_t
-base: $s_base_properties_t
```
aravindksg, line 192: do we need to remove this as well? My thinking was if issue is only with `ZES_MAX_RAS_ERROR_CATEGORY_COUNT`, we just need to remove the last 3 ras error categories?

jandres742: @aravindksg it is better to do a clean revert, instead of pieces of it. These were part of the same commit where the new RAS errors were introduced, so assumption is that they are somehow related. If they were not, then they should have gone originally on different commits.

jandres742 (approved):

bmyates:
Removing these functions and enums is also a backwards-compatibility break.   How can we be certain we aren't breaking a different user/application by removing this?

Issue #81, opened 2023-02-23
should L0 have in-order queues?


jandres742:
L0 has had from the beginning only out-of-order queues, and layers above, like SYCL, have implemented in-order queues by using synchronization primitives like events and barriers.

Although using event or barriers is functionally correct, performance may vary on different generations of the same devices depending on the method used. One way to fix this is to have layers above like SYCL to implement device-dependent in-order queues, however, other more practical option, is to let the L0 driver implementation to directly implement the in-order queue as efficient as possible.

are there other scenarios where in-order queues are desired? or are there any cons on adding this to the spec?

MichalMrozek:
From my point of view it should.
Driver knowing that it works with in order queue is beneficial and it can apply optimizations that are not possible without this knowledge.
In order execution is also well established standard in the industry, so a lot of applicaitons uses it.
Having efficient driver implementation for this is paramount.

abagusetty:
I concur with @MichalMrozek that in-order can be seen mostly as a defacto. I was mostly thinking to provide a case where the SYCL queue construction is in-order and if the underlying L0 queue objects adopt some form of in-order fashion same as the upstream SYCL queue:  DAG construction by the scheduler, the submission of kernels, kernel-scheduling hints can default to oldest-first, dependency tracking, barriers/fences and their associated book-keeping might get simpler, might enable optimizations. 

In reg to optimizations: I am not certain if it is beneficial for L0 to provide the optimizations or the driver to incorporate these. 

More importantly helps applications which has hundreds of "low-runtime" kernel submissions in-order, math libraries and all the projects that are porting from CUDA to SYCL and adopt SYCL in-order as a the reason are some uses cases. Since these types of kernels are flagged to be in-order, any book-keeping overhead with out-of-order can be viewed as an additional overhead.

pboudier09:
we can have a flag either per queue or per device. logically, it should be per queue, since it would give most flexibility to the application.
in practice, it may introduce extra complexity, so having it global per device would be easier for the implementation. If an application wants "out of order", it would need to use separate queues.

MichalMrozek:
In Level Zero entity that needs to ensure in-order is command list , not command queue.
Hence we need command list property to enforce in order start & completion of appended commands.

PR #165, opened 2023-06-26
Add support for in-order lists

jandres742:
Resolves: #81

### DIFF
--- a/scripts/core/cmdlist.yml
+++ b/scripts/core/cmdlist.yml
@@ -30,6 +30,14 @@ etors:
             command list should be optimized for submission to a single command queue and device engine.
             driver **must** disable any implicit optimizations for distributing work across multiple engines.
             this flag should be used when applications want full control over multi-engine submission and scheduling.
+    - name: IN_ORDER
+      version: "1.7"
+      desc: |
+            commands appended to this command list are executed in-order, with driver implementation
+            enforcing dependencies between them. Application is not required to have the signal event
+            of a given command being the wait event of the next to define an in-order list, and
+            application is allowed to pass signal and wait events to each appended command to implement
+            more complex dependency graphs. Cannot be combined with $X_COMMAND_LIST_FLAG_RELAXED_ORDERING.
 --- #--------------------------------------------------------------------------
 type: struct
 desc: "Command List descriptor"


jandres742:
@wdamon-intel , @michalmrozek, @pboudier09: Please review.

scripts/core/cmdlist.yml
```diff
@@ -30,6 +30,14 @@ etors:
...
             driver **must** disable any implicit optimizations for distributing work across multiple engines.
             this flag should be used when applications want full control over multi-engine submission and scheduling.
+    - name: IN_ORDER
+      version: "1.7"
+      desc: |
+            commands appended to this command list are executed in-order, with driver implementation
+            enforcing dependencies between them. Application is not required to have the signal event
+            of a given command being the wait event of the next to define an in-order list, and
+            application is allowed to pass signal and wait events to each appended command to implement
+            more complex dependency graphs. Cannot be combined with $X_COMMAND_LIST_FLAG_RELAXED_ORDERING.
```
MichalMrozek, line 40: "
and application is allowed to pass signal and wait events to each appended command to implement more complex dependency graphs."

Do we need to mention that?  Event model doesn't change with in order command list , you can still pass events from other sources, normally spec mentions something that change or is not allowed compared to default model, while here we state that something that is already there doesn't change.

jandres742: thanks @MichalMrozek . We dont necessarily need to mention it, correct, but I expect this to be a common question. Basically, with in-order lists, users dont need to enforce dependencies to have an in-order list, but can they get the signal events, as well as define additional dependencies? so this is just to clarify that.

wdamon-intel (approved):
LGTM

MichalMrozek (commented):
How those flags can be passed to create immediate command lists?
https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=immediate#_CPPv428zeCommandListCreateImmediate19ze_context_handle_t18ze_device_handle_tPK23ze_command_queue_desc_tP24ze_command_list_handle_t

I don't see flags parameter.

jandres742:
> How those flags can be passed to create immediate command lists? https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=immediate#_CPPv428zeCommandListCreateImmediate19ze_context_handle_t18ze_device_handle_tPK23ze_command_queue_desc_tP24ze_command_list_handle_t
> 
> I don't see flags parameter.

Thanks @MichalMrozek . Adding them also to ze_command_queue_flags_t

MichalMrozek (approved):

scripts/core/cmdqueue.yml
```diff
@@ -20,6 +20,14 @@ etors:
             command queue should be optimized for submission to a single device engine.
             driver **must** disable any implicit optimizations for distributing work across multiple engines.
             this flag should be used when applications want full control over multi-engine submission and scheduling.
+    - name: IN_ORDER
```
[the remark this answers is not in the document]
jandres742: good idea @aravindksg . We will add it.

PR #169, opened 2023-06-27
Add support for flexible device hierarchy model

jandres742:
Resolves: #1

### DIFF
--- a/scripts/core/PROG.rst
+++ b/scripts/core/PROG.rst
@@ -1728,17 +1728,17 @@ The following table documents the supported knobs for overriding default functio
 
 %if ver >= 1.7:
 
-+-----------------+-------------------------------------+-----------------+-----------------------------------------------------------------------------------+
-| Category        | Name                                | Values          | Description                                                                       |
-+=================+=====================================+=================+===================================================================================+
-| Device          | ${X}_FLAT_DEVICE_HIERARCHY            | {**0**, 1, 2}   | Defines device hierarchy model exposed by Level Zero driver implementation        |
-+                 +-------------------------------------+-----------------+-----------------------------------------------------------------------------------+
-|                 | ${X}_AFFINITY_MASK                    | list            | Forces driver to only report devices (and sub-devices) as specified by values     |
-+                 +-------------------------------------+-----------------+-----------------------------------------------------------------------------------+
-|                 | ${X}_ENABLE_PCI_ID_DEVICE_ORDER       | {**0**, 1}      | Forces driver to report devices from lowest to highest PCI bus ID                 |
-+-----------------+-------------------------------------+-----------------+-----------------------------------------------------------------------------------+
-| Memory          | ${X}_SHARED_FORCE_DEVICE_ALLOC        | {**0**, 1}      | Forces all shared allocations into device memory                                  |
-+-----------------+-------------------------------------+-----------------+-----------------------------------------------------------------------------------+
++-----------------+-------------------------------------+-----------------------------------+-----------------------------------------------------------------------------------+
+| Category        | Name                                | Values                            | Description                                                                       |
++=================+=====================================+===================================+===================================================================================+
+| Device          | ${X}_FLAT_DEVICE_HIERARCHY            | {**COMPOSITE**, FLAT, COMBINED}   | Defines device hierarchy model exposed by Level Zero driver implementation        |
++                 +-------------------------------------+-----------------------------------+-----------------------------------------------------------------------------------+
+|                 | ${X}_AFFINITY_MASK                    | list                              | Forces driver to only report devices (and sub-devices) as specified by values     |
++                 +-------------------------------------+-----------------------------------+-----------------------------------------------------------------------------------+
+|                 | ${X}_ENABLE_PCI_ID_DEVICE_ORDER       | {**0**, 1}                        | Forces driver to report devices from lowest to highest PCI bus ID                 |
++-----------------+-------------------------------------+-----------------------------------+-----------------------------------------------------------------------------------+
+| Memory          | ${X}_SHARED_FORCE_DEVICE_ALLOC        | {**0**, 1}                        | Forces all shared allocations into device memory                                  |
++-----------------+-------------------------------------+-----------------------------------+-----------------------------------------------------------------------------------+
 
 
 Device Hierarchy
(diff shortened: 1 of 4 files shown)


jandres742:
@wdamon-intel , @michalmrozek: please review

scripts/core/PROG.rst
```diff
@@ -1714,6 +1723,39 @@ The following table documents the supported knobs for overriding default functio
...
+%endif
+
+%if ver >= 1.7:
+
+Device Hierarchy
+~~~~~~~~~~~~~
+
+${X}_FLAT_DEVICE_HIERARCHY allows users to select the device hierarchy model with which the underlying hardware is exposed and the types of devices returned with ${x}DeviceGet.
+
+With a value of `0`, ${x}DeviceGet returns all the devices that do not have a root-device. Traversing the device hierarchy is possible by querying sub-devices with ${x}DeviceGetSubDevices and root-devices with ${x}DeviceGetRootDevice. Driver implementation may perform implicit optimizations to submissions and allocations done in the root-devices.
```
wdamon-intel, line 1749: Recommend named constants for the values [0, 1, 2].

wdamon-intel (commented):
LGTM, mostly, just a few minor comments to address.

scripts/core/PROG.rst
MichalMrozek, line 1749: in mode 0 DeviceGetRootDevice is not allowed.

scripts/core/PROG.rst
```diff
@@ -1736,6 +1792,16 @@ The following examples demonstrate proper usage for a system configuration of tw
 - `1.1, 1.2`: only device 1 is reported as device 0; with its sub-devices 1 and 2 reported as sub-devices 0 and 1, respectively
 - `0.2, 1.3, 1.0, 0.3`: both device 0 and 1 are reported; device 0 reports sub-devices 2 and 3 as sub-devices 0 and 1, respectively; device 1 reports sub-devices 0 and 3 as sub-devices 0 and 1, respectively; the order is unchanged.
 
+%if ver >= 1.7:
+The following examples show the use of different values in the ${X}_AFFINITY_MASK when setting ${X}_FLAT_DEVICE_HIERARCHY to 1, in the
+same system with two devices and four sub-devices.
+
+- `0, 1, 2, 4`: all sub-devices are reported by ${x}DeviceGet (same as default)
+- `0`: only sub-device 0 in the first device is reported
```
MichalMrozek, line 1800: there are no sub devices when FLAT_DEVICE_HIERACHY is  "1"

scripts/core/PROG.rst
jandres742: @wdamon-intel : you mean like define enums? The main thing about enums is that we would have to give names to each enum, and so we would have to come up with names generic enough to be able to be used by any implementation. IMPLICIT_SCALING shouldn't be used, as that is Intel specific. do you have some names in mind?

jandres742: @MichalMrozek 

You mean in mode 0, which is what we have currently, the application cannot query for the root device? could you elaborate on why would that be? this has been a request from some customers some times, to query for the parent device for a sub-device. From point of view of implementation, this doesnt change too much, as the Level Zero Driver implementation already has that info, it would be just a matter of returning it.

scripts/core/PROG.rst
jandres742: thanks @MichalMrozek . I have expanded this section to make it clearer and added some tables showing what is being exposed. It would look something like this when rendered:

image

scripts/core/PROG.rst
jandres742: @wdamon-intel, @MichalMrozek  : what about:

0 -> COMPOSITE
1->  FLAT
2->  COMBINED or EXPANDEDFLAT or FLATWITHROOT (my preference here would be COMBINED).

MichalMrozek: The API was only meant for mode 2, when you start to flatten the hierarchy.
If hierarchy is not flattened (as in mode 0), then getroot is not supported.

That's the whole difference between mode 2 and other modes, it allows this new API.

jandres742: @MichalMrozek : thanks. yes, the API was originally meant for mode 2, but there's no functional or performance reason to not have it available in mode 0, would that be correct? By having it in mode 0 as well, we allow users to have more flexibility and broader set of APIs to use.

MichalMrozek: let's say you are in mode 0 
app did ZE_AFFINITY_MASK=0.0 
app calls getRoot on such device, what is returned?

jandres742: @MichalMrozek it is return nullptr, but if you are in mode 0, and use ZE_AFFINITY_MASK=0, then you can call getRootDevice on any handle obtained with getSubDevices and the the parent device. It is the same for getSubDevices. With MASK=0, getSubDevices returns a list, and with MASK=0.0 getSubDevices return a count of 0. Spec doesnt forbid the use of getSubDevices with MASK=0.0, it just returns the appropriate value.

MichalMrozek: but in mode 1 , we will return all sub devices as root device and getRoot will return nullptr?

jandres742: Added COMPOSITE, FLAT, COMBINED

jandres742:
@MichalMrozek : do you have further comments or think that PR is good to go?

MichalMrozek:
The initial plan was that new api getRoot is only available in mode 2.
Right now it will be enabled in mode 0, so this would be addition to existing model.
We need clear explanation that resources created for sub devices will not work on extracted root devices, unless root device is a part of the context, same with programs, kernels etc.
Right now we allow that resources created on root devices are available in sub devices , even if they are not in the context, so we need to be careful to not have the same level of expectations with new getRoot API.

jandres742:
> The initial plan was that new api getRoot is only available in mode 2. Right now it will be enabled in mode 0, so this would be addition to existing model. We need clear explanation that resources created for sub devices will not work on extracted root devices, unless root device is a part of the context, same with programs, kernels etc. Right now we allow that resources created on root devices are available in sub devices , even if they are not in the context, so we need to be careful to not have the same level of expectations with new getRoot API.

@MichalMrozek : currently, we allow for sub-devices to share the resources from the root device, that's correct, and you are correct, the same cannot be assumed when you have created a resource in a sub-device, and then getRootDevice is called.

So we could add something like this to getRootDevice:

"To allow the device handle returned by getRootDevice access the resources created by the sub-device handle, these resources need to be created with a context containing explicitly both handles.".

jandres742:
> > The initial plan was that new api getRoot is only available in mode 2. Right now it will be enabled in mode 0, so this would be addition to existing model. We need clear explanation that resources created for sub devices will not work on extracted root devices, unless root device is a part of the context, same with programs, kernels etc. Right now we allow that resources created on root devices are available in sub devices , even if they are not in the context, so we need to be careful to not have the same level of expectations with new getRoot API.
> 
> @MichalMrozek : currently, we allow for sub-devices to share the resources from the root device, that's correct, and you are correct, the same cannot be assumed when you have created a resource in a sub-device, and then getRootDevice is called.
> 
> So we could add something like this to getRootDevice:
> 
> "To allow the device handle returned by getRootDevice access the resources created by the sub-device handle, these resources need to be created with a context containing explicitly both handles.".

Added clarification.

MichalMrozek (approved):

PR #173, opened 2023-06-28
Add API to set atomic properties of a shared allocation

jandres742:


jandres742:
@wdamon-intel , @MichalMrozek, @smilczar: please review.

jandres742:
Defined it as experimental just to have some flexibility if we find out we need or want to change something. but I would expect to be promoted to standard w/o issues. We can also add it as standard now if we want, so open to suggestions.

wdamon-intel:
Agree with introducing this as an experimental extension.

scripts/core/memory.yml
```diff
@@ -540,3 +540,87 @@ type: class
...
 name: $xMem
 owner: $xContext
+--- #--------------------------------------------------------------------------
+type: enum
+desc: "atomic access attribute flags"
+class: $xMem
+name: $x_memory_atomic_attr_exp_flags_t
+version: "1.7"
+etors:
+    - name: NONE
```
MichalMrozek, line 550: this is too coarse and prevents atomics both on host & device.
We need separate controls for no atomics on host & device

jandres742: @MichalMrozek : could you elaborate? didnt understand the part of needing to separate. 
These are flags, not enums, so they could be passed as combination. So for instance:

No atomics anywhere: NONE

No atomics on host, but atomics on device: DEVICE_ATOMIC

Atomics on host, but no atomics on device: DEVICE_ATOMIC

Atomics on host and device: HOST_ATOMIC | DEVICE_ATOMIC

because of the last case, dont know if we actually need SYSTEM_ATOMIC.

MichalMrozek: This doesn't sound from the current description.
By default all shared allocations support atomics / concurrent atomics on both host and device.

Now when you use this API with DEVICE_ATOMIC on such shared allocation you tell the driver that atomics may be performed on the device, but it doesn't tell anything that they are prohibited on the host.

I don't see mapping that you explained in comment above, to what this extensions documents in description, I don't see that when you specify DEVICE_ATOMIC, then it means no atomics on the host.

Only NONE flag disallows atomics, but it does it globally.

MichalMrozek: no_atomics - no atomics performed from device & system
no_host_atomics - no atomics performed from host
no_device_atomics - no atomics performed from device
no_system_atomics - no atomics performed concurrently from host & device
host_atomics - host may do the atomics
device_atomics - device may do the atomics
system_atomics - host & device may do concurrent system atomics, requires support from shared USM cap

jandres742: @MichalMrozek thanks. please take a look.

smilczar:
> @wdamon-intel , @MichalMrozek, @smilczar: please review.

Yes, this set of flags clearly covers all combinations of atomic caps, indeed.

note: we also have examples with flags that say 'yes' or 'no' for a given cap on single bit only:
https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=ze_memory_access_cap_flags_t#ze-memory-access-cap-flags-t

scripts/core/memory.yml
```diff
@@ -540,3 +540,93 @@ type: class
...
+type: enum
+desc: "atomic access attribute flags"
+class: $xMem
+name: $x_memory_atomic_attr_exp_flags_t
+version: "1.7"
+etors:
+    - name: NO_ATOMICS
+      desc: "No atomics performed on the pointer"
+    - name: NO_HOST_ATOMICS
+      desc: "Atomics are not performed on the pointer by the host"
```
MichalMrozek, line 553: can we use allowed instead of performed?

jandres742: @MichalMrozek : yes. Please check.

scripts/core/memory.yml
```diff
@@ -540,3 +540,93 @@ type: class
...
+    - name: HOST_ATOMICS
+      desc: "Host atomics on the pointer are allowed"
+    - name: NO_DEVICE_ATOMICS
+      desc: "Device atomics on the pointer are not allowed"
+    - name: DEVICE_ATOMICS
+      desc: "Device atomics on the pointer are allowed"
+    - name: NO_SYSTEM_ATOMICS
+      desc: "Concurrent atomics on the pointer from both host and device are not allowed"
+    - name: SYSTEM_ATOMICS
+      desc: "Concurrent atomics on the pointer from both host and device are allowed. Requires $X_MEMORY_ACCESS_CAP_FLAG_CONCURRENT_ATOMIC returned by $xDeviceGetMemoryAccessProperties."
```
MichalMrozek, line 563: For completness we need the same mention for host & device atomics.
They are also flags in caps and may not be supported

jandres742: @MichalMrozek : please check.

scripts/core/memory.yml
```diff
@@ -540,3 +540,93 @@ type: class
...
+    - type: $x_memory_atomic_attr_exp_flags_t
+      name: attr
+      desc: |
+            [in] Atomic access attributes to set for the specified range.
+            Must be 0 (default) or a valid combination of $x_memory_atomic_attr_exp_flag_t;
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Retrieves the atomic access attributes previously set for a shared allocation"
+class: $xMem
+name: GetAtomicAccessAttributeExp
```
MichalMrozek, line 615: where is input pointer as argument for this function ?

scripts/core/memory.yml
```diff
@@ -540,3 +540,99 @@ type: class
...
+      name: ptr
+      desc: "[in] Pointer to the start of the memory range"
+    - type: size_t
+      name: size
+      desc: "[in] Size in bytes of the memory range"
+    - type: $x_memory_atomic_attr_exp_flags_t
+      name: attr
+      desc: |
+            [in] Atomic access attributes to set for the specified range.
+            Must be 0 (default) or a valid combination of $x_memory_atomic_attr_exp_flag_t;
```
wdamon-intel, line 610: Remove the semi-colon at the end.

scripts/core/memory.yml
```diff
@@ -540,3 +540,99 @@ type: class
...
+      desc: "Concurrent atomics on the pointer from both host and device are allowed. Requires $X_MEMORY_ACCESS_CAP_FLAG_CONCURRENT_ATOMIC returned by $xDeviceGetMemoryAccessProperties."
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Sets atomic access attributes for a shared allocation"
+class: $xMem
+name: SetAtomicAccessAttributeExp
+version: "1.7"
+details:
+    - |
+      If the shared-allocation is owned by multiple devices (i.e. nullptr
```
MichalMrozek, line 572: how this API works if it is called multiple times with different flags?
is latest call replacing all previous calls or is it additive?

MichalMrozek: I vote for replacing previous setting. Makes driver simpler.

jandres742: @MichalMrozek : please see last commit

scripts/core/memory.yml
jandres742: @wdamon-intel please see last commit

MichalMrozek (approved):

wdamon-intel (approved):

PR #182, opened 2023-07-20
Added Firmware APIs

vaidyan1-int:
Added 2 Firmware related APIs based on customer requests
1. zesFirmwareUpdateProgress - Allows user to get firmware update progress 
2. zesFirmwareLogs - Allows to retrieve Firmware logs


scripts/sysman/firmware.yml
```diff
@@ -90,6 +90,49 @@ returns:
...
+details:
+    - "The application may call this function from simultaneous threads."
+    - "The implementation of this function should be lock-free."
+params:
+    - type: $s_firmware_handle_t
+      name: hFirmware
+      desc: "[in] Handle for the component."
+    - type: "uint32_t*"
+      name: pCompletionPercent
+      desc: "[in,out] Pointer to the Completion Percentage of Firmware Update"
```
aravindksg, line 107: should this be just an "[out]" param?

scripts/sysman/firmware.yml
```diff
@@ -90,6 +90,49 @@ returns:
...
+returns:
+    - $X_RESULT_ERROR_INSUFFICIENT_PERMISSIONS:
+        - "User does not have permissions to perform this operation"
```
aravindksg, line 110: does this require sudo permissions? if not, do we need this return?

scripts/sysman/firmware.yml
```diff
@@ -90,6 +90,49 @@ returns:
...
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Collect Firmware Logs"
+version: "1.7"
+class: $sFirmware
+name: Logs
```
aravindksg, line 116: name could be more descriptive about what kind of FW log this is? i.e, if this is an error log, then name could be "GetErrorLog" ?

wdamon-intel:
@vaidyan1-int please create an issue to associate with this PR.

aravindksg:
adding reviewers: @saik-intel @sanilkumar0 @wdamon-intel @jandres742 @joshuaranjan @jitendrasharma1989

scripts/sysman/firmware.yml
jandres742: this should be out, as there are no requirements to set pCount to a given value (like 0, for zeDeviceGet and alike).

scripts/sysman/firmware.yml
```diff
@@ -90,6 +90,49 @@ returns:
     - $X_RESULT_ERROR_INSUFFICIENT_PERMISSIONS:
         - "User does not have permissions to perform this operation."
 --- #--------------------------------------------------------------------------
+type: function
+desc: "Find Firmware Flash Progress"
+version: "1.7"
+class: $sFirmware
+name: UpdateProgress
```
jandres742, line 97: this is a getter or a setter? I think it is a getter, so, could you add the verb explicitly? like GetProgress ? this would answer also the question from @aravindksg below, if this requires sudo (usually getters dont need sudo).

scripts/sysman/firmware.yml
```diff
@@ -90,6 +90,49 @@ returns:
...
+    - "The caller must provide memory for Firmware log."
+    - "The application may call this function from simultaneous threads."
+    - "The implementation of this function should be lock-free."
+params:
+    - type: $s_firmware_handle_t
+      name: hFirmware
+      desc: "[in] Handle for the component."
+    - type: "size_t*"
+      name: pSize
+      desc: "[in,out] size of firmware log"
```
jandres742, line 128: please specify what happens if user passes a size that is lesser or greater than available.

scripts/sysman/firmware.yml
vaidyan1-int: This is a pointer to the Output value. Will this still be marked as [out] param?  My understanding - pointer to this output value to be provided by app.

scripts/sysman/firmware.yml
vaidyan1-int: Firmware update need sudo permissions

scripts/sysman/firmware.yml
vaidyan1-int: Initial requirement to get AMC FW log. Keeping it generic to expand the scope in future to collect even GPU FW logs

scripts/sysman/firmware.yml
vaidyan1-int: Agree - this is a getter. Make sense to change to GetProgress

scripts/sysman/firmware.yml
vaidyan1-int: Logs will be filled only for the allocated buffer.

scripts/sysman/firmware.yml
joshuaranjan: If possible, we can have the API specific to Flash progress, something like zesFirmwareGetFlashProgress

scripts/sysman/firmware.yml
joshuaranjan: > Initial requirement to get AMC FW log. Keeping it generic to expand the scope in future to collect even GPU FW logs

So please share if there should be a parameter to indicate which FW log this is returning ?
Also zesFirmare"Get"Log could be used

scripts/sysman/firmware.yml
```diff
@@ -90,6 +90,49 @@ returns:
     - $X_RESULT_ERROR_INSUFFICIENT_PERMISSIONS:
         - "User does not have permissions to perform this operation."
 --- #--------------------------------------------------------------------------
+type: function
```
jitendrasharma1989, line 93: @vaidyan1-int I have a query regarding this(Not sure if its worth to discuss, but just want to highlight so that we dont miss any corner case).
Suppose user first called zesFirmwareFlash, and then user called this UpdateProgress.

But as zesFirmwareFlash is not aware that UpdateProgress might be called. So the moment zesFirmwareFlash is called it will immediately flash the firmware. Now if user called UpdateProgress , then i think progress percentage would in max cases start with percentage as greater than 0%.
So,  is it okay?

scripts/sysman/firmware.yml
jitendrasharma1989: @vaidyan1-int Query:
For one firmware handle could we retrieve multiple type of logs? If yes then we might need to add some logType parameter in https://spec.oneapi.io/level-zero/latest/sysman/api.html#_CPPv425zes_firmware_properties_t, and accordingly we need to pass on additional argument to API which will tell about type of log.

scripts/sysman/firmware.yml
jitendrasharma1989: Also could you please tell what is the use case of this API UpdateProgress?
Would user call this UpdateProgress in a loop to get continous update?

Isnt it possible to enhance zesFirmwareFlash API by adding one more parameter named pCompletionPercent?

aravindksg: @jitendrasharma1989 I'd prefer not to combine existing FirmwareFlash API with completion percentage as one is write operation and other is read. Better to decouple

scripts/sysman/firmware.yml
aravindksg: @vaidyan1-int but this is a getter right? not for fw update, but just to gather progress?

scripts/sysman/firmware.yml
jitendrasharma1989: @aravindksg, I am still trying to understand how user would use this updateProgress API?
Is it like from one thread user would invoke a blocking call updateProgress ()
And in another thread user would invoke zesFirmwareFlash(). If that is the case then updateProgress could tell about firmware update progress nicely. But if that is the case then may be we need to add some timeout parameter in updateProgress  API. This timeout parameter would help in letting Sysman know that, some firmware flash command is about to come in this timeout.

aravindksg: so FirmwareFlash is a non-blocking call, the intention of this API is to allow user to immediately call back in same thread for progress update after calling zesFirmwareFlash. So user can call this in a loop at app level. We could add pseudo code in prog guide for this.

scripts/sysman/firmware.yml
aravindksg: @jitendrasharma1989 yeah that's a good point, we can have API name generic such as FirmwareGetLog and have log Type enum as an input

scripts/sysman/firmware.yml
```diff
@@ -90,6 +90,49 @@ returns:
...
+      name: hFirmware
+      desc: "[in] Handle for the component."
+    - type: "size_t*"
+      name: pSize
+      desc: "[in,out] size of firmware log"
+    - type: "char*"
+      name: pFirmwareLog
+      desc: "[in,out][optional] pointer to null-terminated string of the log."  
+returns:
+    - $X_RESULT_ERROR_INSUFFICIENT_PERMISSIONS:
```
aravindksg, line 133: this is also getter API, so would we need this?

scripts/sysman/firmware.yml
jandres742: up to you. Whether we can return the whole log or just partial for size less than available. But we need to specify what happens. 

this

> Logs will be filled only for the allocated buffer.

is what we usually do in other APIs.

scripts/sysman/firmware.yml
saik-intel: @aravindksg   the earlier GPU platforms , FirmwareFlash is blocking call and app need to wait till its done. Is it going to change ?

scripts/sysman/firmware.yml
saik-intel: i agree with @aravindksg  firmwareupdate progress just showing percentage of firmware update done

scripts/sysman/firmware.yml
aravindksg: @saik-intel not seeing anywhere in spec where we have mentioned FirmwareFlash is blocking call. Generally though the APIs are non-blocking unless we have some timeout param

scripts/sysman/firmware.yml
vaidyan1-int: API name changed based on feedback

scripts/sysman/firmware.yml
vaidyan1-int: API name changed

scripts/sysman/firmware.yml
```diff
@@ -90,6 +90,43 @@ returns:
...
+      desc: "[in] Handle for the component."
+    - type: "uint32_t*"
+      name: pCompletionPercent
+      desc: "[in,out] Pointer to the Completion Percentage of Firmware Update"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Get Firmware Logs"
+version: "1.8"
+class: $sFirmware
+name: GetLog
```
aravindksg, line 113: "GetErrorLog"?

wdamon-intel: Probably should keep it generic for flexibility going forward. @jandres742 what do you think?

scripts/sysman/firmware.yml
vaidyan1-int: > @vaidyan1-int Query: For one firmware handle could we retrieve multiple type of logs? If yes then we might need to add some logType parameter in https://spec.oneapi.io/level-zero/latest/sysman/api.html#_CPPv425zes_firmware_properties_t, and accordingly we need to pass on additional argument to API which will tell about type of log.

Firmware logs are pure console logs used for debug purposes. They do not differentiate further like OS logs. That's the reason to keep this simple and not planning to add additional enum to identify log type

scripts/sysman/firmware.yml
```diff
@@ -90,6 +90,43 @@ returns:
...
+params:
+    - type: $s_firmware_handle_t
+      name: hFirmware
+      desc: "[in] Handle for the component."
+    - type: "uint32_t*"
+      name: pCompletionPercent
+      desc: "[in,out] Pointer to the Completion Percentage of Firmware Update"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Get Firmware Logs"
```
aravindksg, line 110: desc to be updated as well to "Get Firmware Console Logs" ?

vaidyan1-int: Done

scripts/sysman/firmware.yml
jandres742: Unless we are purposedly defining this only for ErrorLogs, better to keep it generic.

scripts/sysman/firmware.yml
jandres742: if FirmwareFlash is not blocking, then we should add that to the documentation. I would have expected it to be blocking, and if not, then documentation to mention how user can confirm completion (which might be using getFlashProgress.

scripts/sysman/firmware.yml
```diff
@@ -90,6 +90,43 @@ returns:
...
+      name: pCompletionPercent
+      desc: "[in,out] Pointer to the Completion Percentage of Firmware Update"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Get Firmware Console Logs"
+version: "1.8"
+class: $sFirmware
+name: GetConsoleLogs
+details:
+    - "The caller can pass nullptr for pFirmwareLog when querying only for size."
```
jandres742, line 115: ".. caller may pass ..."

jandres742: also, pSize must be zero when querying for size, right?

vaidyan1-int: Done.

scripts/sysman/firmware.yml
vaidyan1-int: it is generic - zesFirmwareConsoleLogs()

scripts/sysman/firmware.yml
vaidyan1-int: Updated description to include this

aravindksg (approved):

scripts/sysman/firmware.yml
```diff
@@ -76,6 +76,7 @@ name: Flash
 details:
     - "The application may call this function from simultaneous threads."
     - "The implementation of this function should be lock-free."
+    - "The application may call zesFirmwareGetFlashProgress() to get completion status."
```
jandres742, line 79: please expand that this is not a blocking call.

vaidyan1-int: Made this non blocking explicit. Let me know

jandres742 (approved):

PR #183, opened 2023-07-20
Added Memory Page Offline Metrics

vaidyan1-int:
Extended mem_stat_t register to get additional metrics related to Memory page offline.
This is based on a customer requirement.


scripts/sysman/memory.yml
```diff
@@ -132,6 +132,21 @@ members:
       desc: "[out] The total allocatable memory in bytes (can be less than the `physicalSize` member of $s_mem_properties_t)"
 --- #--------------------------------------------------------------------------
 type: struct
+desc: "Extension properties for Memory State"
+version: "1.7"
+class: $sMemory
+name: $s_mem_ext_state_t
```
aravindksg, line 138: should be "mem_state_ext_t" ?

wdamon-intel:
@vaidyan1-int please create an issue to associate with this PR.

scripts/sysman/memory.yml
```diff
@@ -132,6 +132,21 @@ members:
...
+details:
+    - "Memory Page Offline Metrics"
+members:
+    - type: uint32_t
+      name: memoryPageOffline
+      desc: "[out] Returns the number of Memory Pages Offline"
+    - type: uint32_t
+      name: maxMemoryPageOffline
```
joshuaranjan, line 146: Please share how this could be used by the Application.

scripts/sysman/memory.yml
```diff
@@ -132,6 +132,21 @@ members:
...
 type: struct
+desc: "Extension properties for Memory State"
+version: "1.7"
+class: $sMemory
+name: $s_mem_ext_state_t
+details:
+    - "Memory Page Offline Metrics"
+members:
+    - type: uint32_t
+      name: memoryPageOffline
```
joshuaranjan, line 143: Can we have this counter as part of a more granular RAS error category.
Something like ZES_RAS_ERROR_CAT_MEMORY_ECC_ERRORS ?

scripts/sysman/memory.yml
```diff
@@ -132,6 +132,21 @@ members:
       desc: "[out] The total allocatable memory in bytes (can be less than the `physicalSize` member of $s_mem_properties_t)"
 --- #--------------------------------------------------------------------------
 type: struct
+desc: "Extension properties for Memory State"
+version: "1.7"
+class: $sMemory
+name: $s_mem_ext_state_t
+details:
+    - "Memory Page Offline Metrics"
```
joshuaranjan, line 140: Please share if we should consider adding Page size as well ?

scripts/sysman/memory.yml
```diff
@@ -132,6 +132,21 @@ members:
       desc: "[out] The total allocatable memory in bytes (can be less than the `physicalSize` member of $s_mem_properties_t)"
 --- #--------------------------------------------------------------------------
 type: struct
+desc: "Extension properties for Memory State"
```
joshuaranjan, line 135: Please share if we would be adding Enable / Disable of page offlining feature in a new API ?

saik-intel:
@vaidyan1-int   for page offline does first GSC report Memory HEALTH as BAD/DEGRADED and user reboot machine and it will update the offline pages info ?

scripts/sysman/memory.yml
joshuaranjan: ECC has to be disabled to disable this feature

scripts/sysman/memory.yml
vaidyan1-int: Memory page offline is not RAS error. It is a SW functionality to handle ECC errors. That's the reason to keep this separate from RAS

scripts/sysman/memory.yml
vaidyan1-int: Nope. Pagesize might be platform and memory config dependent which we can publish as specs.

scripts/sysman/memory.yml
vaidyan1-int: Yes, ECC disable will disable this feature

vaidyan1-int:
> @vaidyan1-int for page offline does first GSC report Memory HEALTH as BAD/DEGRADED and user reboot machine and it will update the offline pages info ?

The details of the Memory Health reporting based on number of offlined pages is addressed in the SAS

scripts/sysman/memory.yml
saik-intel: @aravindksg  @vaidyan1-int  @sanilkumar0   i think page size is important to update to UMD, it might be different per GPU going ahead ? Isn't it ?

scripts/sysman/memory.yml
saik-intel: > Memory page offline is not RAS error. It is a SW functionality to handle ECC errors. That's the reason to keep this separate from RAS

agreed

scripts/sysman/memory.yml
vaidyan1-int: UMD doesnt need to know this information. Page allocation is handled by KMD and the application need not know any insights into number of offlined pages or page size

scripts/sysman/memory.yml
vaidyan1-int: fixed

aravindksg: @wdamon-intel @jandres742 would you prefer `exp` extension here?

jandres742: @aravindksg : yes please.

vaidyan1-int: Updated the structure with exp. Also added the new files for adding extension.

scripts/sysman/memory.yml
vaidyan1-int: Updated the structure with exp. Also added the new files for adding extension.

scripts/sysman/common.yml
```diff
@@ -211,6 +211,10 @@ etors:
     - name: DEVICE_UUID
       desc: $s_uuid_t
       version: "1.7"
+    - name: MEMORY_EXP_STATE
```
aravindksg, line 214: should be `MEMORY_STATE_EXP`? @wdamon-intel any guidance on this?

scripts/sysman/memState.yml
```diff
@@ -0,0 +1,43 @@
...
+type: struct
+desc: "Extension properties for Memory State"
+version: "1.8"
+class: $sMemory
+name: $s_mem_state_exp_t
+base: $s_base_state_t
+members:
+    - type: uint32_t
+      name: memoryPageOffline
+      desc: "[out] Returns the number of Memory Pages Offline"
```
aravindksg, line 37: to clarify desc- "returns number of memory pages currently offline"?

scripts/sysman/memState.yml
```diff
@@ -0,0 +1,43 @@
...
+    - type: uint32_t
+      name: maxMemoryPageOffline
+      desc: "[out] Returns the Allowed Memory Pages Offline"
```
aravindksg, line 40: "returns max allowed memory pages offline" ?

scripts/sysman/common.yml
wdamon-intel: @aravindksg `MEMORY_STATE_EXP` is correct

joshuaranjan, line 214: Could we name the structure type more specific to the feature.
Something like MEMORY_PAGE_OFFLINE_STATE_EXP.

This would help to extend MEMORY_STATE in the future.

vaidyan1-int: Fixed in common.yml

scripts/sysman/common.yml
```diff
@@ -211,6 +211,10 @@ etors:
     - name: DEVICE_UUID
       desc: $s_uuid_t
       version: "1.7"
+    - name: MEMORY_STATE_EXP
```
wdamon-intel, line 214: Per the suggestion by @joshuaranjan in the conversation, perhaps a more specific name such as `MEMORY_PAGE_OFFLINE_STATE_EXP` would be preferable, for both the name and the structure type.

vaidyan1-int: Would this create a new data structure instead of the extension to the original mem_state structure?

aravindksg: @vaidyan1-int no, I think ask here was to just rename your extension proposal to be more specific. (something like `zes_memory_page_offline_state_exp`) rather than simply `zes_memory_state_exp`. We can always expose multiple extensions via the pNext in existing zes_mem_state_t. if we keep struct names as specific now, it would help to differentiate any new extensions we might need to add here

vaidyan1-int: Got it. I understand the intent. Will this also initiate change to the yml file name to memPageOfflineState.yml ?

aravindksg: yes, would be good to rename file as well. thanks!

vaidyan1-int: Done 👍

scripts/sysman/EXT_Exp_MemState.rst
```diff
@@ -0,0 +1,32 @@
+<%
```
aravindksg, line 1: could you please rename this file as well to "EXT_Exp_MemPageOfflineState.rst"?

vaidyan1-int: Done.

aravindksg (approved):

PR #191, opened 2023-08-03
Move engine activity extension to separate extension file

aravindksg:


aravindksg:
@jandres742 @wdamon-intel @sanilkumar0 @saik-intel @joshuaranjan

jandres742 (approved):

scripts/sysman/engineActivity.yml
```diff
@@ -0,0 +1,64 @@
...
+#
+# See YaML.md for syntax definition
+#
+--- #--------------------------------------------------------------------------
+type: header
+desc: "Intel $OneApi Level-Zero Sysman Extension APIs for Engine Activity"
+version: "1.7"
+--- #--------------------------------------------------------------------------
+type: macro
+desc: "Power Limits Extension Name"
```
joshuaranjan, line 14: We may have to rename this

aravindksg: thanks @joshuaranjan fixed

joshuaranjan (approved):

aravindksg:
@jandres742 @joshuaranjan @sanilkumar0 please re-review. have added pseudo code and also clarifications on return value and pCount description

scripts/sysman/EXT_EngineActivity.rst
```diff
@@ -0,0 +1,76 @@
...
+    // utilization for all VFs and PF as shown below
+
+    ${s}_engine_properties_t engineProps;
+    ${s}_engine_ext_properties_t engineExtProps;
+    engineProps.stype = ZES_STRUCTURE_TYPE_ENGINE_PROPERTIES;
+    engineProps.pNext = &engineExtProps;
+    engineExtProps.stype = ZES_STRUCTURE_TYPE_ENGINE_EXT_PROPERTIES;
+
+    ${s}EngineGetProperties(hEngine, &engineProps);
+    uint32_t numVF = engineExtProps.countOfVirtualFunctionInstance;
```
joshuaranjan, line 54: This seems unused.
So do we want to mention it here ?

scripts/sysman/EXT_EngineActivity.rst
```diff
@@ -0,0 +1,76 @@
...
+Intended use case of the feature is to query utilization stats from Physical Function (PF) interface. However, the extension
+may return error code $X_RESULT_ERROR_UNSUPPORTED_FEATURE if an environment does not support it.
+
+The following pseudo-code demonstrates a sequence for obtaining the engine activity for all Virtual Functions from Physical Function environment:
+
+.. parsed-literal::
+
+    // Query for engine properties and gather number of virtual function instances by passing extension structure.
+    // Number of VFs is per engine handle and equates to number of VFs actively using
+    // the resource represented by the engine handle.
```
joshuaranjan, line 43: Is it "by the engine" OR "engine handle" ?
handle objects would be different right ?

scripts/sysman/EXT_EngineActivity.rst
```diff
@@ -0,0 +1,76 @@
...
+    // Gather engine stats for all VFs and PF.
+    // Gather two snapshots of engine activity and calculate percentage
+    // utilization per formula in ${s}_engine_stats_t
+    ${s}_engine_stats_t* engineStats0 = (${s}_engine_stats_t*) allocate(engineStatCount * sizeof(${s}_engine_stats_t));
+    {s}EngineGetActivityExt(hEngine, &engineStatCount, engineStats0);
+    sleep(1);
+    ${s}_engine_stats_t* engineStats1 = (${s}_engine_stats_t*) allocate(engineStatCount * sizeof(${s}_engine_stats_t));
+    {s}EngineGetActivityExt(hEngine, &engineStatCount, engineStats1);
+
+    for (uint32_t vf = 0; vf < engineStatCount; ++vf) {
```
joshuaranjan, line 69: here vf means PF as well right (at the last index) ?

scripts/sysman/EXT_EngineActivity.rst
```diff
@@ -0,0 +1,76 @@
...
+        // Ensure there was activity before calculation
+        if (engineStats0[vf].activeTime == 0 && engineStats1[vf].activeTime == 0) {
```
joshuaranjan, line 71: Clarification: Do we need this check, since the utilization would show 0%.
Also since the active time is monotonically increasing, 0% utilization may mean that activeTime at 2 snapshots are same (may not be necessarily 0)

scripts/sysman/EXT_EngineActivity.rst
```diff
@@ -0,0 +1,76 @@
...
+* Structures
+
+    * ${s}_engine_ext_properties_t
+
+Engine Activity
+~~~~~~~~~~~~~~~~~~
+
+This extension provides capability to user to gather the activity stats for each Virtual Function (VF) associated with engine group.
+Intended use case of the feature is to query utilization stats from Physical Function (PF) interface. However, the extension
+may return error code $X_RESULT_ERROR_UNSUPPORTED_FEATURE if an environment does not support it.
```
joshuaranjan, line 35: Just thinking: Should we have ther error as unsupported feature - this may mean that this feature would be supported in the future. 
OR insufficient permissions, which may mean the VF does not have the privileges to retrieve the necessary information.

scripts/sysman/engineActivity.yml
```diff
@@ -0,0 +1,73 @@
...
+version: "1.7"
+class: $sEngine
+name: $s_engine_ext_properties_t
+base: $s_base_properties_t
+members:
+    - type: uint32_t
+      name: "countOfVirtualFunctionInstance"
+      desc: |
+            [out] Number of Virtual Function(VF) instances associated with engine to monitor the global utilization of hardware across all Virtual Function from a Physical Function (PF) instance.
+            These global and VF-by-VF views should provide engine group and individual engine level granularity.
```
joshuaranjan, line 39: global includes PF usage as well right ? 
So should we mention only "VF views" here.

scripts/sysman/engineActivity.yml
```diff
@@ -0,0 +1,73 @@
...
+      desc: |
+            [in,out] Pointer to the number of VF engine stats descriptors.
+              - if count is zero, the driver shall update the value with the total number of engine stats available.
+              - if count is greater than the total number of engine stats available, the driver shall update the value with the correct number of engine stats available.
+              - The count returned is the sum of number of VF instances currently available and the PF instance.
+    - type: $s_engine_stats_t*
+      name: pStats
+      desc: |
+            [in,out][optional][range(0, *pCount)] array of engine group activity counters.
+              - if count is less than the total number of engine stats available, then driver shall only retrieve that number of stats.
```
joshuaranjan, line 68: How do we handle PF activity in this case.
Would PF activity be still be at the last index ?

scripts/sysman/engineActivity.yml
```diff
@@ -0,0 +1,73 @@
...
+              - the implementation shall populate the vector with user provided pCount number of VF engine stats followed by engine stat for PF.
+              - If VF has no activity value to report, then implementation shall use value of 0 to denote no activity.
```
joshuaranjan, line 70: Clarification:
Would activity time be "0" ? If yes we could move this to $s_engine_stats_t description.
But I think the spec mentions that activity time is monotonically increasing. Please share if this will change.

scripts/sysman/engineActivity.yml
joshuaranjan: @sanilkumar0 could we have PF activity at index 0 ?

scripts/sysman/EXT_EngineActivity.rst
aravindksg: I could also add error return for insufficient permissions here that way implementation can choose to return whichever is more suitable

scripts/sysman/EXT_EngineActivity.rst
aravindksg: good point, shall remove this

scripts/sysman/EXT_EngineActivity.rst
aravindksg: yes, will change iterator name

scripts/sysman/EXT_EngineActivity.rst
aravindksg: yep, will remove check

scripts/sysman/engineActivity.yml
aravindksg: yeah good point, this count represents number of VFs actively using particular resource. so I can remove global wording. does that seem fine @sanilkumar0 ?

scripts/sysman/engineActivity.yml
aravindksg: will fix the index

scripts/sysman/engineActivity.yml
aravindksg: good point, i'll remove the lines from here and add note in the extension docs about interpreting zero activity

scripts/sysman/EXT_EngineActivity.rst
aravindksg: so insufficient permission is actually when "user does not have permissions to perform an operation". in this case it's not about user but VF instance that can't read the value due to environment. so i think might be better to use unsupported feature. adding description to that effect.

scripts/sysman/EXT_EngineActivity.rst
aravindksg: fixed

scripts/sysman/EXT_EngineActivity.rst
aravindksg: fixed

scripts/sysman/EXT_EngineActivity.rst
aravindksg: fixed

scripts/sysman/engineActivity.yml
aravindksg: fixed

scripts/sysman/engineActivity.yml
aravindksg: fixed

scripts/sysman/engineActivity.yml
aravindksg: fixed

scripts/sysman/engineActivity.yml
```diff
@@ -0,0 +1,72 @@
...
+      desc: |
+            [out] Number of Virtual Function(VF) instances associated with engine to monitor the utilization of hardware across all Virtual Function from a Physical Function (PF) instance.
+            These VF-by-VF views should provide engine group and individual engine level granularity.
+            This count represents the number of VF instances that are actively using the resource represented by the engine handle.
+details:
+    - "This structure may be passed to $sEngineGetProperties by having the pNext member of $s_engine_properties_t point at this struct."
+    - "Used for SRIOV per Virtual Function device utilization by $s_engine_group_t"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Get activity stats for each Virtual Function (VF) associated with engine group."
```
joshuaranjan, line 46: We may have to add "PF" here

joshuaranjan:
lgtm
Thanks !

PR #199, opened 2023-08-23
Clarify metric streamer desc member descriptions

### LINKED ISSUE #135: missing details on zet_metric_streamer_desc_t out parameters
zet_metric_streamer_desc_t https://spec.oneapi.io/level-zero/latest/tools/api.html#_CPPv426zet_metric_streamer_desc_t has a couple parameters that work as input but can be updated by the driver implementation. However, there is no clear description on the expected output. Updated text should say something like: 

**uint32_t notifyEveryNReports**
[in,out] number of collected reports after which notification event will be signaled. **The driver shall updated the value with the actual closest number it supports.** 

uint32_t samplingPeriod
[in,out] streamer sampling period in nanoseconds. **The driver shall updated the value with the actual closest number it supports.**

wdamon-intel:
Resolves #135

### DIFF
--- a/scripts/tools/metric.yml
+++ b/scripts/tools/metric.yml
@@ -289,10 +289,14 @@ base: $t_base_desc_t
 members:
     - type: uint32_t
       name: "notifyEveryNReports"
-      desc: "[in,out] number of collected reports after which notification event will be signalled"
+      desc: >
+            [in,out] number of collected reports after which notification event will be signaled.
+            The driver shall update this field with the closest supported value during $tMetricStreamerOpen.
     - type: uint32_t
       name: "samplingPeriod"
-      desc: "[in,out] streamer sampling period in nanoseconds"
+      desc: >
+            [in,out] streamer sampling period in nanoseconds.
+            The driver shall update this field with the closest supported value during $tMetricStreamerOpen.
 --- #--------------------------------------------------------------------------
 type: function
 desc: "Opens metric streamer for a device."


wdamon-intel:
@matcabral please review

matcabral (approved):

scripts/tools/metric.yml
```diff
@@ -289,10 +289,14 @@ base: $t_base_desc_t
 members:
     - type: uint32_t
       name: "notifyEveryNReports"
-      desc: "[in,out] number of collected reports after which notification event will be signalled"
+      desc: >
+            [in,out] number of collected reports after which notification event will be signaled.
+            The driver shall update this field with the closest supported value during $tMetricStreamerOpen.
```
jandres742, line 294: @matcabral : what do we mean by "closest"? also, MetricStreamerOpen returns a handle, so to which value are we referring to?

matcabral: @jandres742  please see https://github.com/oneapi-src/level-zero-spec/issues/135 

"closest" == most approximate to what the user requested but cannot exactly be respected for the current hardware so the implementation has the flexibility to chose something else respecting user suggestion.

wdamon-intel: Modified the wording per @jandres742 concern. @matcabral @jandres742 please review.

matcabral (approved):

jandres742 (approved):

Issue #201, opened 2023-08-28
[Tools][Metrics] Add APIs to support Configurable Metrics


joshuaranjan:
The request is to define APIs to allow the applications to choose some of the attributes of metrics like result type (whether normalized / averaged / ..) rather than using pre-defined attributes.

The implementation could define the level of configurability of the metrics

joshuaranjan:
PR is at https://github.com/oneapi-src/level-zero-spec/pull/204

@wdamon-intel @matcabral @jandres742 @aravindksg

PR #204, opened 2023-08-29
Add support for programmable metrics

joshuaranjan:
Resolves: #201

### DIFF
--- a/scripts/tools/EXT_Exp_MetricProgrammable.rst
+++ b/scripts/tools/EXT_Exp_MetricProgrammable.rst
@@ -1,0 +1,120 @@
+<%
+import re
+from templates import helper as th
+%><%
+    OneApi=tags['$OneApi']
+    x=tags['$x']
+    X=x.upper()
+    t=tags['$t']
+    T=t.upper()
+%>
+:orphan:
+
+.. _ZET_experimental_programmable_metric:
+
+==========================================
+Metric Programmable Experimental Extension
+==========================================
+
+API
+----
+* Enumerations
+
+    * ${t}_metric_programmable_param_type_exp_t
+
+* Structures
+
+    * ${t}_metric_programmable_exp_properties_t
+    * ${t}_metric_programmable_param_exp_desc_t
+    * ${t}_metric_programmable_param_value_exp_t
+    * ${t}_metric_programmable_param_value_exp_desc_t
+    * ${t}_value_exp_desc_t
+    * ${t}_value_uint64_range_exp_t
+
+* Functions
+
+    * ${t}MetricProgrammableGetExp
+    * ${t}MetricProgrammableGetPropertiesExp
+    * ${t}MetricProgrammableGetParamDescriptionExp
+    * ${t}MetricProgrammableGetParamValueDescriptionExp
+    * ${t}MetricCreateFromProgrammableExp
+    * ${t}MetricGroupCreateExp
+    * ${t}MetricGroupAddMetricExp
+    * ${t}MetricDestroyExp
+    * ${t}MetricGroupDestroyExp
+
+Metric Programmable
+~~~~~~~~~~~~~~~~~~~
+
+Application can use ${t}MetricGroupGet to enumerate the list of metric groups and the metrics available for collection.
+However Applications may also want to collect custom metrics which are not listed as part of the enumeration.
+This is made possible by making a list of programmable metrics available to the application, so that it choose different parameter values for the preparing custom metrics.
+The following pseudo-code demonstrates how programmable metrics could be enumerated and custom values could be set to the metric parameters.
+
+Sample Code
+------------
+
+The following pseudo-code demonstrates how programmable metrics could be enumerated and custom values could be set to the metric parameters.
+
+.. parsed-literal::
+
+    ${t}_metric_handle_t * metricHandles;
+    uint32_t metricHandleCount = 0;
+    ${t}_metric_group_handle_t metricGroup;
+    
+    // Query and Get metric programmable handles
+    uint32_t programmableCount = 0;
+    ${t}MetricProgrammableGetExp(device, &programmableCount, nullptr);
+    ${t}_metric_programmable_exp_handle_t * metricProgrammableHandles = allocate(sizeof(${t}_metric_programmable_exp_handle_t) * programmableCount);
+    ${t}MetricProgrammableGetExp(device, &programmableCount, metricProgrammableHandles);
+
+    // Create metrics from metric programmable handles
+    for(uint32_t i = 0; i < programmableCount; i++){
+    
+        ${t}_metric_programmable_exp_handle_t * programmableHandle = metricProgrammableHandles[i];
+        ${t}_metric_programmable_exp_properties_t programmableProperties{};
+        ${t}MetricProgrammableGetPropertiesExp(programmableHandle, &programmableProperties);
+        
+        // Choose programmable handle of interest
+        if(strcmp(programmableProperties.name, "EU_ACTIVE" ) == 0){
+            
+            // Get Parameter descriptions
+            ${t}_metric_programmable_param_exp_desc_t * paramDescription = allocate (sizeof(${t}_metric_programmable_param_exp_desc_t) * programmableProperties.parameterCount);
+            ${t}MetricProgrammableGetParamDescriptionExp(programmableHandle, programmableProperties.parameterCount, paramDescription);
+            
+            // Get Parameter Value descriptions for the 0th parameter
+            ${t}_metric_programmable_param_value_exp_desc_t * valueDesc = allocate(sizeof(${t}_metric_programmable_param_value_exp_desc_t) * paramDescription[0].valueDescriptionCount);
+            ${t}MetricProgrammableGetParamValueDescriptionExp(programmableHandle, 0, paramDescription[0].valueDescriptionCount, valueDesc);
+            
+            // Setting value for the 0th parameter only
+            ${t}_metric_programmable_param_value_exp_t parameterValue;
+            parameterValue.value = valueDesc[0].valueDesc.ui64;
+            
+            // Create Metric
+            ${t}_metric_handle_t metricHandle{};
+            char metricName[ZET_MAX_METRIC_NAME] = "eu_active_minimum";
+            char metricDescription[ZET_MAX_METRIC_DESCRIPTION] = "eu_active_minimum_desc";
+            ${t}MetricCreateFromProgrammableExp(programmableHandle, &parameterValue, 1, metricName, metricDescription, &metricHandleCount, null_ptr);
+            metricHandles = allocate(sizeof(${t}_metric_handle_t) * metricHandleCount);
+            ${t}MetricCreateFromProgrammableExp(programmableHandle, &parameterValue, 1, metricName, metricDescription, &metricHandleCount, metricHandles);
+        }
+        
+        //Create Metric Group from metrics
+        uint32_t handlesSize = metricHandlesUsed.size();
+        char metricGroupName[ZET_MAX_METRIC_GROUP_NAME] = "eu_active";
+        char metricGroupDescription[ZET_MAX_METRIC_GROUP_DESCRIPTION] = "eu_active_desc";
+        ${t}MetricGroupCreateExp(device, metricGroupName, metricGroupDescription, &metricGroup);
+        ${t}MetricGroupAddMetricExp(metricGroup, &metricHandlesUsed[0], null_ptr, null_ptr);
+    }
+    
+    //Activate Metric group
+    //Collect Metric group using available sampling types
+    
+    //Cleanup
+    ${t}MetricGroupDestroyExp(metricGroup);
+    
+    for(uint32_t j = 0; j < metricHandleCount; j++){
+        ${t}MetricDestroyExp(metricHandles[j]);
+    }
+
+
--- a/scripts/tools/EXT_Exp_MetricProgrammable.rst
+++ b/scripts/tools/EXT_Exp_MetricProgrammable.rst
@@ -106,7 +106,7 @@ The following pseudo-code demonstrates how programmable metrics could be enumera
     ${t}_metric_group_handle_t metricGroupHandle{};
     char metricGroupName[ZET_MAX_METRIC_GROUP_NAME] = "eu_active";
     char metricGroupDescription[ZET_MAX_METRIC_GROUP_DESCRIPTION] = "eu_active_desc";
-    ${t}MetricGroupCreateExp(device, metricGroupName, metricGroupDescription, &metricGroup);
+    ${t}MetricGroupCreateExp(device, metricGroupName, metricGroupDescription, ZET_METRIC_GROUP_SAMPLING_TYPE_FLAG_TIME_BASED, &metricGroup);
     ${t}MetricGroupAddMetricExp(metricGroup, &metricHandles[0], null_ptr, null_ptr);
     ${t}MetricGroupFinalizeExp(metricGroup);
 
--- a/scripts/tools/EXT_Exp_MetricProgrammable.rst
+++ b/scripts/tools/EXT_Exp_MetricProgrammable.rst
@@ -86,7 +86,7 @@ The following pseudo-code demonstrates how programmable metrics could be enumera
             ${t}MetricProgrammableGetParamInfoExp(programmableHandle, programmableProperties.parameterCount, paramInfo);
 
             // Get Parameter Value info for the 0th parameter
-            ${t}_metric_programmable_param_value_info_exp_t * paramValueInfo = allocate(sizeof(${t}_metric_programmable_param_value_info_exp_t) * paramInfo[0].paramValueInforiptionCount);
+            ${t}_metric_programmable_param_value_info_exp_t * paramValueInfo = allocate(sizeof(${t}_metric_programmable_param_value_info_exp_t) * paramInfo[0].valueInfoCount);
             ${t}MetricProgrammableGetParamValueInfoExp(programmableHandle, 0, paramInfo[0].valueInfoCount, paramValueInfo);
 
             // Setting value for the 0th parameter
@@ -94,16 +94,15 @@ The following pseudo-code demonstrates how programmable metrics could be enumera
             parameterValue.value = paramValueInfo[0].valueInfo.ui64;
 
             // Create Metric
-            ${t}_metric_handle_t metricHandle;
             char metricName[ZET_MAX_METRIC_NAME] = "eu_active_minimum";
             char metricDescription[ZET_MAX_METRIC_DESCRIPTION] = "eu_active_minimum_desc";
             ${t}MetricCreateFromProgrammableExp(programmableHandle, &parameterValue, 1, metricName, metricDescription, &metricHandleCount, null_ptr);
-            metricHandles = allocate(sizeof(${t}_metric_handle_t) * metricHandleCount);
+            ${t}_metric_handle_t * metricHandles = allocate(sizeof(${t}_metric_handle_t) * metricHandleCount);
             ${t}MetricCreateFromProgrammableExp(programmableHandle, &parameterValue, 1, metricName, metricDescription, &metricHandleCount, metricHandles);
         }
 
         //Create Metric Group from metrics
-        uint32_t handlesSize = metricHandlesUsed.size();
+        ${t}_metric_group_handle_t metricGroupHandle{};
         char metricGroupName[ZET_MAX_METRIC_GROUP_NAME] = "eu_active";
         char metricGroupDescription[ZET_MAX_METRIC_GROUP_DESCRIPTION] = "eu_active_desc";
         ${t}MetricGroupCreateExp(device, metricGroupName, metricGroupDescription, &metricGroup);
--- a/scripts/tools/EXT_Exp_MetricProgrammable.rst
+++ b/scripts/tools/EXT_Exp_MetricProgrammable.rst
@@ -43,7 +43,7 @@ API
     * ${t}MetricGroupCreateExp
     * ${t}MetricGroupAddMetricExp
     * ${t}MetricGroupRemoveMetricExp
-    * ${t}MetricGroupFinalizeExp
+    * ${t}MetricGroupCloseExp
     * ${t}MetricDestroyExp
     * ${t}MetricGroupDestroyExp
 
@@ -108,7 +108,7 @@ The following pseudo-code demonstrates how programmable metrics could be enumera
     char metricGroupDescription[ZET_MAX_METRIC_GROUP_DESCRIPTION] = "eu_active_desc";
     ${t}MetricGroupCreateExp(device, metricGroupName, metricGroupDescription, ZET_METRIC_GROUP_SAMPLING_TYPE_FLAG_TIME_BASED, &metricGroup);
     ${t}MetricGroupAddMetricExp(metricGroup, &metricHandles[0], null_ptr, null_ptr);
-    ${t}MetricGroupFinalizeExp(metricGroup);
+    ${t}MetricGroupCloseExp(metricGroup);
 
     //Activate Metric group
     //Collect Metric group using available sampling types
--- a/scripts/tools/EXT_Exp_MetricProgrammable.rst
+++ b/scripts/tools/EXT_Exp_MetricProgrammable.rst
@@ -50,7 +50,7 @@ API
 Metric Programmable
 ~~~~~~~~~~~~~~~~~~~
 
-Application can use ${t}MetricGroupGet to enumerate the list of metric groups and the metrics available for collection.
+Application can use ${t}MetricGroupGet to enumerate the list of metric groups and ${t}MetricGet for the metrics available for collection.
 However Applications may also want to collect custom metrics which are not listed as part of the enumeration.
 This is made possible by making a list of programmable metrics available to the application, so that it may choose different parameter values for preparing custom metrics.
 
@@ -100,14 +100,15 @@ The following pseudo-code demonstrates how programmable metrics could be enumera
             ${t}_metric_handle_t * metricHandles = allocate(sizeof(${t}_metric_handle_t) * metricHandleCount);
             ${t}MetricCreateFromProgrammableExp(programmableHandle, &parameterValue, 1, metricName, metricDescription, &metricHandleCount, metricHandles);
         }
-
-        //Create Metric Group from metrics
-        ${t}_metric_group_handle_t metricGroupHandle{};
-        char metricGroupName[ZET_MAX_METRIC_GROUP_NAME] = "eu_active";
-        char metricGroupDescription[ZET_MAX_METRIC_GROUP_DESCRIPTION] = "eu_active_desc";
-        ${t}MetricGroupCreateExp(device, metricGroupName, metricGroupDescription, &metricGroup);
-        ${t}MetricGroupAddMetricExp(metricGroup, &metricHandlesUsed[0], null_ptr, null_ptr);
     }
+
+    //Create Metric Group from metrics
+    ${t}_metric_group_handle_t metricGroupHandle{};
+    char metricGroupName[ZET_MAX_METRIC_GROUP_NAME] = "eu_active";
+    char metricGroupDescription[ZET_MAX_METRIC_GROUP_DESCRIPTION] = "eu_active_desc";
+    ${t}MetricGroupCreateExp(device, metricGroupName, metricGroupDescription, &metricGroup);
+    ${t}MetricGroupAddMetricExp(metricGroup, &metricHandles[0], null_ptr, null_ptr);
+    ${t}MetricGroupFinalizeExp(metricGroup);
 
     //Activate Metric group
     //Collect Metric group using available sampling types
--- a/scripts/tools/EXT_Exp_MetricProgrammable.rst
+++ b/scripts/tools/EXT_Exp_MetricProgrammable.rst
@@ -40,6 +40,7 @@ API
     * ${t}MetricCreateFromProgrammableExp
     * ${t}MetricGroupCreateExp
     * ${t}MetricGroupAddMetricExp
+    * ${t}MetricGroupRemoveMetricExp
     * ${t}MetricDestroyExp
     * ${t}MetricGroupDestroyExp
 
@@ -49,7 +50,6 @@ Metric Programmable
 Application can use ${t}MetricGroupGet to enumerate the list of metric groups and the metrics available for collection.
 However Applications may also want to collect custom metrics which are not listed as part of the enumeration.
 This is made possible by making a list of programmable metrics available to the application, so that it choose different parameter values for the preparing custom metrics.
-The following pseudo-code demonstrates how programmable metrics could be enumerated and custom values could be set to the metric parameters.
 
 Sample Code
 ------------
@@ -58,10 +58,10 @@ The following pseudo-code demonstrates how programmable metrics could be enumera
 
 .. parsed-literal::
 
-    ${t}_metric_handle_t * metricHandles;
+    ${t}_metric_handle_t * metricHandles = null_ptr;
     uint32_t metricHandleCount = 0;
     ${t}_metric_group_handle_t metricGroup;
-    
+
     // Query and Get metric programmable handles
     uint32_t programmableCount = 0;
     ${t}MetricProgrammableGetExp(device, &programmableCount, nullptr);
@@ -70,26 +70,26 @@ The following pseudo-code demonstrates how programmable metrics could be enumera
 
     // Create metrics from metric programmable handles
     for(uint32_t i = 0; i < programmableCount; i++){
-    
+
         ${t}_metric_programmable_exp_handle_t * programmableHandle = metricProgrammableHandles[i];
         ${t}_metric_programmable_exp_properties_t programmableProperties{};
         ${t}MetricProgrammableGetPropertiesExp(programmableHandle, &programmableProperties);
-        
+
         // Choose programmable handle of interest
         if(strcmp(programmableProperties.name, "EU_ACTIVE" ) == 0){
-            
+
             // Get Parameter descriptions
             ${t}_metric_programmable_param_exp_desc_t * paramDescription = allocate (sizeof(${t}_metric_programmable_param_exp_desc_t) * programmableProperties.parameterCount);
             ${t}MetricProgrammableGetParamDescriptionExp(programmableHandle, programmableProperties.parameterCount, paramDescription);
-            
+
             // Get Parameter Value descriptions for the 0th parameter
             ${t}_metric_programmable_param_value_exp_desc_t * valueDesc = allocate(sizeof(${t}_metric_programmable_param_value_exp_desc_t) * paramDescription[0].valueDescriptionCount);
             ${t}MetricProgrammableGetParamValueDescriptionExp(programmableHandle, 0, paramDescription[0].valueDescriptionCount, valueDesc);
-            
-            // Setting value for the 0th parameter only
+
+            // Setting value for the 0th parameter
             ${t}_metric_programmable_param_value_exp_t parameterValue;
             parameterValue.value = valueDesc[0].valueDesc.ui64;
-            
+
             // Create Metric
             ${t}_metric_handle_t metricHandle{};
             char metricName[ZET_MAX_METRIC_NAME] = "eu_active_minimum";
@@ -98,7 +98,7 @@ The following pseudo-code demonstrates how programmable metrics could be enumera
             metricHandles = allocate(sizeof(${t}_metric_handle_t) * metricHandleCount);
             ${t}MetricCreateFromProgrammableExp(programmableHandle, &parameterValue, 1, metricName, metricDescription, &metricHandleCount, metricHandles);
         }
-        
+
         //Create Metric Group from metrics
         uint32_t handlesSize = metricHandlesUsed.size();
         char metricGroupName[ZET_MAX_METRIC_GROUP_NAME] = "eu_active";
@@ -106,13 +106,13 @@ The following pseudo-code demonstrates how programmable metrics could be enumera
         ${t}MetricGroupCreateExp(device, metricGroupName, metricGroupDescription, &metricGroup);
         ${t}MetricGroupAddMetricExp(metricGroup, &metricHandlesUsed[0], null_ptr, null_ptr);
     }
-    
+
     //Activate Metric group
     //Collect Metric group using available sampling types
-    
+
     //Cleanup
     ${t}MetricGroupDestroyExp(metricGroup);
-    
+
     for(uint32_t j = 0; j < metricHandleCount; j++){
         ${t}MetricDestroyExp(metricHandles[j]);
     }
--- a/scripts/tools/EXT_Exp_MetricProgrammable.rst
+++ b/scripts/tools/EXT_Exp_MetricProgrammable.rst
@@ -21,22 +21,24 @@ API
 * Enumerations
 
     * ${t}_metric_programmable_param_type_exp_t
+    * ${t}_value_info_type_exp_t
+    * ${t}_metric_programmable_exp_version_t
 
 * Structures
 
     * ${t}_metric_programmable_exp_properties_t
-    * ${t}_metric_programmable_param_exp_desc_t
+    * ${t}_metric_programmable_param_info_exp_t
     * ${t}_metric_programmable_param_value_exp_t
-    * ${t}_metric_programmable_param_value_exp_desc_t
-    * ${t}_value_exp_desc_t
+    * ${t}_metric_programmable_param_value_info_exp_t
+    * ${t}_value_info_exp_t
     * ${t}_value_uint64_range_exp_t
 
 * Functions
 
     * ${t}MetricProgrammableGetExp
     * ${t}MetricProgrammableGetPropertiesExp
-    * ${t}MetricProgrammableGetParamDescriptionExp
-    * ${t}MetricProgrammableGetParamValueDescriptionExp
+    * ${t}MetricProgrammableGetParamInfoExp
+    * ${t}MetricProgrammableGetParamValueInfoExp
     * ${t}MetricCreateFromProgrammableExp
     * ${t}MetricGroupCreateExp
     * ${t}MetricGroupAddMetricExp
@@ -49,7 +51,7 @@ Metric Programmable
 
 Application can use ${t}MetricGroupGet to enumerate the list of metric groups and the metrics available for collection.
 However Applications may also want to collect custom metrics which are not listed as part of the enumeration.
-This is made possible by making a list of programmable metrics available to the application, so that it choose different parameter values for the preparing custom metrics.
+This is made possible by making a list of programmable metrics available to the application, so that it may choose different parameter values for preparing custom metrics.
 
 Sample Code
 ------------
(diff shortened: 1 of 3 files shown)


wdamon-intel (commented):
@joshuaranjan just a few comments/questions. See the review comments for details.

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,475 @@
...
+members:
+    - type: $t_value_exp_desc_t
+      name: "valueDesc"
+      desc: "[out] description of the parameter value"
+--- #--------------------------------------------------------------------------
+type: struct
+desc: "Metric Programmable parameter value"
+version: "1.7"
+class: $tMetricProgrammable
+name: $t_metric_programmable_param_value_exp_t
```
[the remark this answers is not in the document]
joshuaranjan: Currently there are base types only for properties and descriptors .
But felt the need to add structure_type and pNext here  for future extension.

Please let me know if I should be creating a new base type ?

scripts/tools/EXT_Exp_MetricProgrammable.rst
```diff
@@ -0,0 +1,120 @@
...
+
+==========================================
+Metric Programmable Experimental Extension
+==========================================
+
+API
+----
+* Enumerations
+
+    * ${t}_metric_programmable_param_type_exp_t
```
aravindksg, line 23: you would also need to enumerate `${t}_metric_programmable_exp_version_t`

scripts/tools/EXT_Exp_MetricProgrammable.rst
```diff
@@ -0,0 +1,120 @@
...
+    * ${t}_metric_programmable_param_value_exp_desc_t
+    * ${t}_value_exp_desc_t
+    * ${t}_value_uint64_range_exp_t
+
+* Functions
+
+    * ${t}MetricProgrammableGetExp
+    * ${t}MetricProgrammableGetPropertiesExp
+    * ${t}MetricProgrammableGetParamDescriptionExp
+    * ${t}MetricProgrammableGetParamValueDescriptionExp
```
aravindksg, line 39: generally `desc` are input args, so here it seems we are providing desc to gather output right? is there different way to name this API/ output structures ?

scripts/tools/EXT_Exp_MetricProgrammable.rst
```diff
@@ -0,0 +1,120 @@
...
+    * ${t}MetricCreateFromProgrammableExp
+    * ${t}MetricGroupCreateExp
+    * ${t}MetricGroupAddMetricExp
+    * ${t}MetricDestroyExp
```
aravindksg, line 44: we don't have a equivalent MetricCreateExp here. so should this be named differently?

scripts/tools/EXT_Exp_MetricProgrammable.rst
```diff
@@ -0,0 +1,120 @@
...
+    * ${t}MetricGroupDestroyExp
+
+Metric Programmable
+~~~~~~~~~~~~~~~~~~~
+
+Application can use ${t}MetricGroupGet to enumerate the list of metric groups and the metrics available for collection.
+However Applications may also want to collect custom metrics which are not listed as part of the enumeration.
+This is made possible by making a list of programmable metrics available to the application, so that it choose different parameter values for the preparing custom metrics.
```
aravindksg, line 52: suggest reword "so that it may choose different parameter values for preparing custom metrics"

scripts/tools/EXT_Exp_MetricProgrammable.rst
```diff
@@ -0,0 +1,120 @@
...
+The following pseudo-code demonstrates how programmable metrics could be enumerated and custom values could be set to the metric parameters.
+
```
aravindksg: statement repeated below

scripts/tools/EXT_Exp_MetricProgrammable.rst
```diff
@@ -0,0 +1,120 @@
...
+    // Query and Get metric programmable handles
+    uint32_t programmableCount = 0;
+    ${t}MetricProgrammableGetExp(device, &programmableCount, nullptr);
+    ${t}_metric_programmable_exp_handle_t * metricProgrammableHandles = allocate(sizeof(${t}_metric_programmable_exp_handle_t) * programmableCount);
+    ${t}MetricProgrammableGetExp(device, &programmableCount, metricProgrammableHandles);
+
+    // Create metrics from metric programmable handles
+    for(uint32_t i = 0; i < programmableCount; i++){
+    
+        ${t}_metric_programmable_exp_handle_t * programmableHandle = metricProgrammableHandles[i];
```
aravindksg, line 74: is it metric_programmable_exp_handle_t programmableHandle = metricProgrammableHandles[i] here?
(i.e, once you index into the vector, you are getting the handle right?)

scripts/tools/EXT_Exp_MetricProgrammable.rst
```diff
@@ -0,0 +1,120 @@
...
+    }
+    
+    //Activate Metric group
+    //Collect Metric group using available sampling types
+    
+    //Cleanup
+    ${t}MetricGroupDestroyExp(metricGroup);
+    
+    for(uint32_t j = 0; j < metricHandleCount; j++){
+        ${t}MetricDestroyExp(metricHandles[j]);
```
aravindksg, line 117: MetricDestroyFromProgrammable (or some other name) for symmetry?

scripts/tools/common.yml
```diff
@@ -180,7 +200,7 @@ members:
       desc: "[out] 32-bit unsigned-integer"
     - type: uint64_t
       name: "ui64"
-      desc: "[out] 32-bit unsigned-integer"
+      desc: "[out] 64-bit unsigned-integer"
```
aravindksg, line 203: this correction is unrelated to this PR, I think would be good to make this correction as separate PR anyway since we can merge that quickly while we revise the proposal here?

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,445 @@
...
+      desc: "Produces normalization in percent using raw_metric * 100 / cycles / instance_count."
+    - name: NORMALIZATION_AVERAGE
+      desc: "Produces normalization using raw_metric / instance_count."
+    - name: NORMALIZATION_RATE
+      desc: "Produces normalization average using raw_metric / timestamp."
+--- #--------------------------------------------------------------------------
+type: enum
+desc: "Supported value description types"
+version: "1.7"
+name: $t_value_type_exp_desc_t
```
aravindksg, line 130: descriptors generally contain members, since this is enum this should be renamed

scripts/tools/EXT_Exp_MetricProgrammable.rst
```diff
@@ -0,0 +1,120 @@
+<%
+import re
+from templates import helper as th
+%><%
+    OneApi=tags['$OneApi']
+    x=tags['$x']
+    X=x.upper()
```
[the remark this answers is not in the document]
joshuaranjan: yes

scripts/tools/EXT_Exp_MetricProgrammable.rst
joshuaranjan: Ok .. Will add that.
Thanks

scripts/tools/EXT_Exp_MetricProgrammable.rst
joshuaranjan: Let me check this.
But please share if there is a guideline about the names.

scripts/tools/EXT_Exp_MetricProgrammable.rst
joshuaranjan: The idea is that metrics could be created from different mechansims but destroy to be applicable for all created metrics.
So did not want to have the creation method added to the destroy name.
(Just a note - Destroy would be able to detect how it was created internally)

scripts/tools/EXT_Exp_MetricProgrammable.rst
joshuaranjan: Ok.

scripts/tools/EXT_Exp_MetricProgrammable.rst
joshuaranjan: Ok. 
(this was fixed in the  last commit)

scripts/tools/EXT_Exp_MetricProgrammable.rst
joshuaranjan: You are correct.
Will fix

scripts/tools/EXT_Exp_MetricProgrammable.rst
joshuaranjan: Like mentioned above, 
"The idea is that metrics could be created from different mechansims but destroy to be applicable for all created metrics.
So did not want to have the creation method added to the destroy name."

scripts/tools/common.yml
joshuaranjan: ok

scripts/tools/metricProgrammable.yml
joshuaranjan: Ok.
Please share if there is a guideline for it.
Thanks

scripts/tools/EXT_Exp_MetricProgrammable.rst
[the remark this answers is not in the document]
joshuaranjan: We use $X and $x in the .yml. 
So please share if "X=" and "x=" can be removed from .rst

scripts/tools/metricProgrammable.yml
joshuaranjan: renamed to info

scripts/tools/common.yml
joshuaranjan: done

scripts/tools/EXT_Exp_MetricProgrammable.rst
joshuaranjan: fixed

scripts/tools/EXT_Exp_MetricProgrammable.rst
joshuaranjan: fixed

scripts/tools/EXT_Exp_MetricProgrammable.rst
joshuaranjan: renamed to info

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,475 @@
...
+# Copyright (C) 2023 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+#
+# See YaML.md for syntax definition
+#
+--- #--------------------------------------------------------------------------
+type: header
+desc: "Intel $OneApi Level-Zero Tool Experimental Extension for Programmable Metrics"
+version: "1.7"
```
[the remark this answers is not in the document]
joshuaranjan: done

scripts/tools/common.yml
```diff
@@ -125,6 +125,26 @@ etors:
       desc: $t_metric_calculate_exp_desc_t
       version: "1.7"
       value: "0x00010002"
+    - name: METRIC_PROGRAMMABLE_EXP_PROPERTIES
+      desc: $t_metric_programmable_exp_properties_t
+      version: "1.7"
```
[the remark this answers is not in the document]
joshuaranjan: done

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+      name: "parameterCount"
+      desc: "[out] number of parameters in the programmable"
+    - type: "$t_metric_group_sampling_type_flags_t"
+      name: "samplingType"
+      desc: |
+            [out] metric sampling type.
+            returns a combination of $t_metric_group_sampling_type_flags_t.
+    - type: uint32_t
+      name: "sourceId"
+      desc: "[out] unique metric source identifier(within platform)to identify the HW block where the metric is collected."
```
RobertKrzemien, line 108: Can the API be extended in the future? If not, consider adding `disaggregation mode` and `query type mask` as new members

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+params:
+    - type: $t_metric_group_handle_t 
+      name: metricGroupHandle
+      desc: "[in] Handle of the metric group to destroy"
+returns:
+    - $X_RESULT_ERROR_INVALID_ARGUMENT:
+        - "If trying to destroy a pre-defined metric group"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Destroy a created metric."
```
RobertKrzemien, line 479:
```suggestion
desc: "Destroy a metric created using $tMetricCreateFromProgrammableExp."
```

scripts/tools/metricProgrammable.yml
joshuaranjan: >Can the API be extended in the future?
Thanks. yes, the api can be extended in the future.

There are a few changes being considered for tracer/event mode of collection.
We could recheck these properties when those changes are integrated as well.

scripts/tools/metricProgrammable.yml
joshuaranjan: Ok

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+desc: "Add a metric handle to the metric group handle created using $tMetricGroupCreateExp."
+version: "1.8"
+class: $tMetricGroup
+name: AddMetricExp
+decl: static
+details:
+    - "Reasons for failing to add the metric could be queried using pErrorString"
+params:
+    - type: $t_metric_group_handle_t 
+      name: metricGroupHandle
```
joshuaranjan, line 411: hMetricGroup

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+      desc: "[in] Handle of the metric group"
+    - type: const $t_metric_handle_t*
```
joshuaranjan, line 413: pointer to be removed, since only 1 metric is added

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+      name: phMetric
```
joshuaranjan, line 414: hMetric

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+desc: "Remove a metric from the metric group handle created using $tMetricGroupCreateExp."
+version: "1.8"
+class: $tMetricGroup
+name: RemoveMetricExp
+decl: static
+details:
+    - "Remove an already added metric handle from the metric group"
+params:
+    - type: $t_metric_group_handle_t
+      name: metricGroupHandle
```
joshuaranjan, line 440: hMetricGroup

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+      desc: "[in] Handle of the metric group"
+    - type: const $t_metric_handle_t
```
joshuaranjan, line 442: remove const

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Destroy a metric group created using $tMetricGroupCreateExp."
+version: "1.8"
+class: $tMetricGroup
+name: DestroyExp
+decl: static
+params:
+    - type: $t_metric_group_handle_t 
+      name: metricGroupHandle
```
joshuaranjan, line 472: hMetricGroup

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+class: $tMetricGroup
+name: FinalizeExp
+decl: static
+details:
+    - "Finalize the metric group."
+    - "This is a necessary step for the created metric group, before activation."
+    - "Add / Remove of metrics is possible after Finalize. However, a call to $tMetricGroupFinalizeExp is necessary after modifying the metric group."
+params:
+    - type: $t_metric_group_handle_t
+      name: metricGroupHandle
```
joshuaranjan, line 461: hMetricGroup

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Destroy a metric created using $tMetricCreateFromProgrammableExp."
+version: "1.8"
+class: $tMetric
+name: DestroyExp
+decl: static
+params:
+    - type: $t_metric_handle_t
+      name: metricHandle
```
joshuaranjan, line 486: hMetric

scripts/tools/metricProgrammable.yml
```diff
@@ -476,7 +476,7 @@ returns:
         - "If trying to destroy a pre-defined metric group"
 --- #--------------------------------------------------------------------------
 type: function
-desc: "Destroy a created metric."
+desc: "Destroy a metric created using $tMetricCreateFromProgrammableExp."
```
aravindksg, line 479: I think you had mentioned in an earlier comment you intended destroy to be applicable for all created metrics. but here you have updated desc to say this is applicable only for those created from ProgrammableExp. If so, I think it would be better to rename destroy API to reflect that

scripts/tools/EXT_Exp_MetricProgrammable.rst
```diff
@@ -0,0 +1,123 @@
...
+
+.. _ZET_experimental_programmable_metric:
+
+==========================================
+Metric Programmable Experimental Extension
+==========================================
+
+API
+----
+* Enumerations
```
matcabral, line 21: where are new handle types defined?

scripts/tools/EXT_Exp_MetricProgrammable.rst
```diff
@@ -0,0 +1,123 @@
...
+
+* Structures
+
+    * ${t}_metric_programmable_exp_properties_t
+    * ${t}_metric_programmable_param_info_exp_t
+    * ${t}_metric_programmable_param_value_exp_t
+    * ${t}_metric_programmable_param_value_info_exp_t
+    * ${t}_value_info_exp_t
+    * ${t}_value_uint64_range_exp_t
+
```
matcabral, line 35: zet_metric_programmable_param_description_t ? needed by zetMetricProgrammableGetParamInfo()

scripts/tools/metricProgrammable.yml
joshuaranjan: >I think you had mentioned in an earlier comment you intended destroy to be applicable for all created metrics.

You are correct. This API is meant to destroy all created metrics.
I had updated the description here, since we have only zetMetricCreateFromProgrammableExp to create metrics now.
In the future, when we have more apis to create metrics, I thought we could update the description accordingly.

However, if it is confusing, I can rever the change

scripts/tools/metricProgrammable.yml
joshuaranjan: done

scripts/tools/metricProgrammable.yml
joshuaranjan: done

scripts/tools/metricProgrammable.yml
joshuaranjan: done

scripts/tools/metricProgrammable.yml
joshuaranjan: done

scripts/tools/metricProgrammable.yml
joshuaranjan: done

scripts/tools/metricProgrammable.yml
joshuaranjan: done

scripts/tools/metricProgrammable.yml
joshuaranjan: done

scripts/tools/EXT_Exp_MetricProgrammable.rst
joshuaranjan: Those are defined in metricProgrammable.yml
https://github.com/oneapi-src/level-zero-spec/pull/204/files#diff-5f98b851645e6662ea9e2b7ac56a0566cd73f10f5bf2fe42b91db481bb4723a0

scripts/tools/EXT_Exp_MetricProgrammable.rst
joshuaranjan: It has been renamed with "info", since description has a different meaning in other places of L0 spec
zet_metric_programmable_param_info_exp_t

scripts/tools/EXT_Exp_MetricProgrammable.rst
```diff
@@ -0,0 +1,123 @@
...
+            // Create Metric
+            ${t}_metric_handle_t metricHandle;
+            char metricName[ZET_MAX_METRIC_NAME] = "eu_active_minimum";
+            char metricDescription[ZET_MAX_METRIC_DESCRIPTION] = "eu_active_minimum_desc";
+            ${t}MetricCreateFromProgrammableExp(programmableHandle, &parameterValue, 1, metricName, metricDescription, &metricHandleCount, null_ptr);
+            metricHandles = allocate(sizeof(${t}_metric_handle_t) * metricHandleCount);
+            ${t}MetricCreateFromProgrammableExp(programmableHandle, &parameterValue, 1, metricName, metricDescription, &metricHandleCount, metricHandles);
+        }
+
+        //Create Metric Group from metrics
```
matcabral, line 105: seems this should be inside the `if(strcmp(programmableProperties.name, "EU_ACTIVE" ) == 0){`

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+details:
+    - "Returns information about the parameters of the metric programmable handle."
+params:
+    - type: $t_metric_programmable_exp_handle_t
+      name: hMetricProgrammable
+      desc: "[in] handle of the metric programmable"
+    - type: "uint32_t*"
+      name: pParameterCount
+      desc: |
+            [in,out] count of the parameters to retrieve parameter info.
```
matcabral, line 305: - missing the behavior of 0 to query the number. 
- What happens if the number is less than the actual available? 

General rules applicable for all the *count [in,out] params in the PR, since I found some comments apply to others:
- passing 0 to get the actual number
- passing a value smaller the available.
- passing a value bigger than available

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+            if value pParameterCount is greater than count of parameters available, then pParameterCount will be updated with count of parameters available.
+    - type: "$t_metric_programmable_param_info_exp_t*"
+      name: pParameterInfo
+      desc: |
+            [out][range(1, *pParameterCount)] array of parameter info.
```
matcabral, line 310: - I saw that **all** APIs that have arrays that are **always** outputs are labeled as [in,out]. @wdamon-intel  would you please clarify why?
- also described as [optional] because you may submit nullptr when querying for the count. 

Above comments applies to all output arrays in the PR.

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+type: function
+desc: "Create metric handles by applying parameter values on the metric programmable handle."
+version: "1.8"
+class: $tMetric
+name: CreateFromProgrammableExp
+decl: static
+details:
+    - "Multiple parameter values could be used to prepare a metric."
+    - "If parameterCount = 0, the default value of the metric programmable would be used for all parameters"
+    - "Parameter values are updated to pParameterValues in the same order as retrieved using $tMetricProgrammableGetParamInfoExp"
```
matcabral, line 348: not clear what the implications of the statement are. Would you please elaborate?

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+    - "If multiple metrics are generated, then implementation defined C string will be post-fixed to metric name and description"
```
matcabral, line 349: - if multiple metrics are created in a single call to the API? 
- "then implementation defined C string will be post-fixed to metric name and description"  does this mean at the user input will not be verbatim respected? why would this only apply to the case of multiple metrics?

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Create metric group handle."
+version: "1.8"
+class: $tMetricGroup
+name: CreateExp
+decl: static
+details:
+    - "Metric handles from multiple domains may be used for a single metric group."
+    - "Metric handles from different sourceIds (refer $t_metric_programmable_exp_properties_t) are not allowed in a single metric group."
```
matcabral, line 386: It seems that both of these comments are more related to when the user adds metrics to a group rather than when creating the group.

scripts/tools/EXT_Exp_MetricProgrammable.rst
```diff
@@ -0,0 +1,123 @@
...
+    * ${t}MetricGroupAddMetricExp
+    * ${t}MetricGroupRemoveMetricExp
+    * ${t}MetricGroupFinalizeExp
+    * ${t}MetricDestroyExp
+    * ${t}MetricGroupDestroyExp
+
+Metric Programmable
+~~~~~~~~~~~~~~~~~~~
+
+Application can use ${t}MetricGroupGet to enumerate the list of metric groups and the metrics available for collection.
```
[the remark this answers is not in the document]
joshuaranjan: Ok .. will update

scripts/tools/EXT_Exp_MetricProgrammable.rst
joshuaranjan: Yes .. the closing braces seem incorrect.
Will fix.
Thanks

scripts/tools/metricProgrammable.yml
joshuaranjan: Here the usage of pParameterCount is slightly different:

Value of parameterCount is part of programmable properties ($t_metric_programmable_exp_properties_t).
So application has that information from `GetPropertiesExp`

Here the pParameterCount is for the application to specify the number of parameters it would like to query "information" for.

Similarly in the other APIs as well.

scripts/tools/metricProgrammable.yml
aravindksg: ok sounds good, latest change looks fine by me

scripts/tools/metricProgrammable.yml
joshuaranjan: Like mentioned in the previous comment, this is not an optional parameter.
Application would pass in the pointer allocated according to pParameterCount

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+params:
+    - type: $t_metric_group_handle_t 
+      name: hMetricGroup
+      desc: "[in] Handle of the metric group"
+    - type: $t_metric_handle_t
+      name: hMetric
+      desc: "[in] Metric to be added to the group."
+    - type: "uint32_t*"
+      name: pErrorStringSize
+      desc: "[in,out][optional] Size of the error string to query, if an error was reported during adding the metric handle."
```
matcabral, line 418: missing something like "mist be 0 if pErrorString == nullptr"  like in 
https://spec.oneapi.io/level-zero/latest/tools/api.html#zetcontextactivatemetricgroups

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+decl: static
+details:
+    - "Returns the properties of the metric programmable."
+params:
+    - type: $t_metric_programmable_exp_handle_t
+      name: hMetricProgrammable
+      desc: "[in] handle of the metric programmable"
+    - type: "$t_metric_programmable_exp_properties_t*"
+      name: pProperties
+      desc: "[out] properties of the metric programmable"
```
matcabral, line 288: multiple functions with no return error codes defined.

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+params:
+    - type: $t_metric_group_handle_t
+      name: hMetricGroup
+      desc: "[in] Handle of the metric group"
+    - type: $t_metric_handle_t
+      name: hMetric
+      desc: "[in] Metric handle to be removed from the metric group."
+returns:
+    - $X_RESULT_ERROR_INVALID_ARGUMENT:
+        - "If trying to remove a metric not previously added to the metric group"
```
matcabral, line 447: how about calling it on an active metric group?

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+class: $tMetricGroup
+name: DestroyExp
+decl: static
+params:
+    - type: $t_metric_group_handle_t 
+      name: hMetricGroup
+      desc: "[in] Handle of the metric group to destroy"
+returns:
+    - $X_RESULT_ERROR_INVALID_ARGUMENT:
+        - "If trying to destroy a pre-defined metric group"
```
matcabral, line 476: - is it allowed to destroy a metric group that contains metrics? if yes, need to explicitly clarify it destroys the metrics 
- what if the metric group is active? 
-

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+class: $tMetric
+name: DestroyExp
+decl: static
+params:
+    - type: $t_metric_handle_t
+      name: hMetric
+      desc: "[in] Handle of the metric to destroy"
+returns:
+    - $X_RESULT_ERROR_INVALID_ARGUMENT:
+        - "If trying to destroy a metric from pre-defined metric group"
```
matcabral, line 490: - what if the metric currently belongs to a metric group?

scripts/tools/metricProgrammable.yml
matcabral: all output arrays that can accept nullptr are described in the spec as optional.

wdamon-intel: @matcabral regarding the `[in,out][optional][range(1, ...)]` on array parameters in APIs such as this, here's a general set of guidelines (note these are not strictly official definitions).

- `in`: the parameter represents an immutable object supplied and managed by the user
- `out`: the parameter represents a mutable object that may be modified and/or managed by the implementation
- `in,out`: the parameter represents an immutable reference to a mutable object; the contents of the referenced object may change as a side effect of the API
- `optional`: the parameter may be `NULL`, the validation layer will not issue a warning or error and the implementation is expected to anticipate this condition.
- `range(x, y)`: the parameter represents a container object with a minimum of _x_ elements and a maximum of _y_ elements, indexable range is [_(x - 1), (y - 1)_]

matcabral: @wdamon-intel Thanks for clarifying. In general I propose to have a common Reference section where all this details are described. This is a source of confusion. 

As a side topic, the fact that the distinction of in and out is made based on the object and not the information/value contained in the object seems contradictory. For instance, IMHO I would expect that "in" is for pieces of data the user feeds to the API for a given purpose and "out" is a piece of data expected as a result of an action of the API. I would expect that params that are "in and out" are for cases in which the user value input is in fact used (as a guidance) and the API updates it to the actual value used (like https://spec.oneapi.io/level-zero/latest/tools/api.html#_CPPv4N26zet_metric_streamer_desc_t14samplingPeriodE).  Labeling [in,out] parameters where the user only inputs the "container" (with no meaningful data for the API) where the API only provides valid output  content should be labeled as [out] (as in this PR). Especially because conflicts with params where there is effectively data going in and out.

scripts/tools/common.yml
```diff
@@ -125,6 +125,26 @@ etors:
...
+      value: "0x00010004"
+    - name: METRIC_PROGRAMMABLE_PARAM_VALUE_EXP
+      desc: $t_metric_programmable_param_value_exp_t
+      version: "1.8"
+      value: "0x00010005"
+    - name: METRIC_PROGRAMMABLE_PARAM_VALUE_INFO_EXP
+      desc: $t_metric_programmable_param_value_info_exp_t
+      version: "1.8"
+      value: "0x00010006"
+    - name: VALUE_INFO_EXP
```
jandres742, line 144: why this one is not prefixed with METRIC_?

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+name: "$t_metric_programmable_exp_handle_t"
+--- #--------------------------------------------------------------------------
+type: struct
+desc: "Metric Programmable properties queried using $tMetricProgrammableGetPropertiesExp"
+version: "1.8"
+class: $tMetricProgrammable
+name: $t_metric_programmable_exp_properties_t
+base: $t_base_properties_t
+members:
+    - type: char
```
jandres742, line 83: all interfaces in L0 have an owner (a driverHandle, a deviceHandle, a metric handle, etc), what is the one for this interface?

scripts/tools/metricProgrammable.yml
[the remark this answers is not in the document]
jandres742: @joshuaranjan : let's remove the stype. Base type for properties and descriptors. if we think that we need to have something expandable for the future, then let's define a property or descriptor parameter  (or this one as such) for CreateFromProgrammableExp.

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+      desc: "[in] Metric handle to be removed from the metric group."
+returns:
+    - $X_RESULT_ERROR_INVALID_ARGUMENT:
+        - "If trying to remove a metric not previously added to the metric group"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Finalizes a created metric group using $tMetricGroupCreateExp, so that it can be activated."
+version: "1.8"
+class: $tMetricGroup
+name: FinalizeExp
```
jandres742, line 453: if the counter API of this is tMetricGroupCreateExp, then this should be called Destroy, following the convention of other APIs in the spec.

jandres742: but I see there's a destroy below. what is finalize and what is destroy? why do we need both? it seems confusing as both refer to tMetricGroupCreateExp

>"Finalizes a created metric group using $tMetricGroupCreateExp
> "Destroy a metric group created using $tMetricGroupCreateExp."

Create goes with Destroy. Finalize seems to need a Start or Begin interface.

scripts/tools/common.yml
joshuaranjan: Existing "Value" related structures / enums does not seems to be prefixed with metrics.
For example: t_value_type_t

So thought of having the value_info_x structure without a metric prefix.

scripts/tools/metricProgrammable.yml
joshuaranjan: Could you please clarify more.

t_metric_programmable_exp_properties_t is a structure and not an interface/API

scripts/tools/metricProgrammable.yml
joshuaranjan: >if the counter API of this is tMetricGroupCreateExp,
This API is NOT a counter API for MetricGroupCreate.
This API is for finalizing (stop configuring by using the ADD / REMOVE APIs) the metric group.

Just a note: We have MetricGroupDestroy as well.
Thanks

joshuaranjan: > it seems confusing as both refer to tMetricGroupCreateExp

Ok . Let me update the description to be more precise about its usage.

scripts/tools/metricProgrammable.yml
matcabral: OK, so only one case for in an out is needed. Clear

scripts/tools/metricProgrammable.yml
matcabral: @joshuaranjan based on the above rule, this (and all other arrays) would need to be in,out

scripts/tools/metricProgrammable.yml
matcabral: I think the explanation should be more explicit.  something like
"tMetricProgrammableGetParamInfoExp() returns a list of parameter's in a defined order. Therefore, the list of values passed in to this API should respect the same order such that the desired parameter is set with the expected value"

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,490 @@
...
+      desc: "[in,out][optional] Size of the error string to query, if an error was reported during adding the metric handle."
+    - type: "char*"
+      name: pErrorString
+      desc: |
+            [in,out][optional][range(0, *pErrorStringSize)] Error string.
+            if *pErrorStringSize is less than the length of the error string available, then driver shall only retrieve that length of error string.
+returns:
+    - $X_RESULT_ERROR_INVALID_ARGUMENT:
+        - "If a Metric handle from a pre-defined metric group is requested to be added"
+    - $X_RESULT_ERROR_UNSUPPORTED_FEATURE:
```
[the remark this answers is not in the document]
matcabral: I mean if the the API was not implemented, implementations should return UNSUPORTED_FEATURE. So, I assume you mean the error if already activated should OBJECT_IN_USE, right?

scripts/tools/metricProgrammable.yml
matcabral: should this description be added?

scripts/tools/metricProgrammable.yml
joshuaranjan: @matcabral 
Yes .. Let me make that change

@wdamon-intel Thanks for the explanation.
If we could document this somewhere (maybe a contributing.md), it would help to be consistent across APIs

scripts/tools/metricProgrammable.yml
jandres742: thanks @joshuaranjan . I still dont understand what Finalize does. could you share pseudo code  or add a section in [scripts/tools/EXT_Exp_MetricProgrammable.rst](https://github.com/oneapi-src/level-zero-spec/pull/204/files#diff-04edfdabb9ef36b6fc56e99cd7e7cb4a1df5867167c10f78cec8b6aec3b1570f) showing how it is used?

scripts/tools/metricProgrammable.yml
joshuaranjan: Done

scripts/tools/metricProgrammable.yml
joshuaranjan: Move it to add api

scripts/tools/metricProgrammable.yml
joshuaranjan: I think default return error codes are added by the infra.

scripts/tools/metricProgrammable.yml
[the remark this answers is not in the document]
joshuaranjan: Yes.

scripts/tools/metricProgrammable.yml
joshuaranjan: Added it

scripts/tools/metricProgrammable.yml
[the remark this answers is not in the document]
joshuaranjan: removed the base type and pNext for this structure

scripts/tools/metricProgrammable.yml
joshuaranjan: updated the explanation for size==0 case

scripts/tools/metricProgrammable.yml
joshuaranjan: Updated all arrays with [in, out].

scripts/tools/metricProgrammable.yml
joshuaranjan: >"then implementation defined C string will be post-fixed to metric name and description" does this mean at the user input will not be verbatim respected? why would this only apply to the case of multiple metrics?

Corrected it. 
Implementations can post-fix C strings to metric names based on the parameter values set.

aravindksg (commented):
+1 changes lgtm

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,513 @@
...
+      name: "sourceId"
+      desc: "[out] unique metric source identifier(within platform)to identify the HW block where the metric is collected."
+--- #--------------------------------------------------------------------------
+type: enum
+desc: "Metric Programmable Parameter types"
+version: "1.8"
+class: $tMetricProgrammable
+name: $t_metric_programmable_param_type_exp_t
+etors:
+    - name: ETOR_DISAGGREGATION
```
wdamon-intel, line 116: Why the `ETOR_` prefix? The scripts will expand the name. For example, `DISAGGREGATION` will expand to `ZET_METRIC_PROGRAMMABLE_PARAM_TYPE_DISAGGREGATION`.

The same comment/question follows for all the etors in this list and in the "`value_info_type`" etors below.

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,513 @@
...
+      desc: "[out] minimum value of the range"
+    - type: uint64_t
+      name: "ui64Max"
+      desc: "[out] max value of the range"
+--- #--------------------------------------------------------------------------
+type: union
+desc: "Union of value information"
+version: "1.8"
+name: $t_value_info_exp_t
+base: $t_base_desc_t
```
wdamon-intel, line 167: Do the scripts allow for deriving a union from a base descriptor type? If so then that is a bug in the scripts.

joshuaranjan: I dont see an error while running.
I removed the base type as well

joshuaranjan:
@wdamon-intel @jandres742 
Could you please have a relook at this PR

scripts/tools/common.yml
```diff
@@ -157,7 +169,7 @@ etors:
     - name: UINT32
       desc: "32-bit unsigned-integer"
     - name: UINT64
-      desc: "64-bit unsigned-integer"
+      desc: "32-bit unsigned-integer"
```
wdamon-intel, line 172: 32 -> 64

wdamon-intel (changes requested):

scripts/tools/common.yml
joshuaranjan: This was fixed in this PR earlier.
But received a comment to revert, since it may not be related to this scope.
Reference : https://github.com/oneapi-src/level-zero-spec/pull/204#discussion_r1310855530

However, I can still go ahead and fix this. Please suggest.

jandres742 (approved):

scripts/tools/common.yml
joshuaranjan: @wdamon-intel 
fixed it.

wdamon-intel (approved):

aravindksg (approved):

matcabral (approved):

scripts/tools/metricProgrammable.yml
```diff
@@ -0,0 +1,512 @@
...
+# Copyright (C) 2023 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+#
+# See YaML.md for syntax definition
+#
+--- #--------------------------------------------------------------------------
+type: header
+desc: "Intel $OneApi Level-Zero Tool Experimental Extension for Programmable Metrics"
+version: "1.8"
```
aravindksg, line 11: @joshuaranjan please change all versions to "1.9"

joshuaranjan: done.

scripts/tools/common.yml
```diff
@@ -125,6 +125,18 @@ etors:
       desc: $t_metric_calculate_exp_desc_t
       version: "1.7"
       value: "0x00010002"
+    - name: METRIC_PROGRAMMABLE_EXP_PROPERTIES
+      desc: $t_metric_programmable_exp_properties_t
+      version: "1.8"
```
aravindksg, line 130: @joshuaranjan version to 1.9 here as well

joshuaranjan: Sorry missed it.
Updated now.

Issue #197, opened 2023-08-21
Remove time units for activeTime and timestamp fields in zes_engine_stats_t


aravindksg:
Proposal here is to remove “microseconds” from activeTIme and timestamp. By obfuscating the units, we can directly use hardware counters that might be counting at frequencies different than microsecs/milliseconds. It also means that when we provide numbers to VFs, customers will not be assuming that these numbers are correlated to wall time. As the description says, the only purpose for these numbers is for a percentage calculation

aravindksg:
@saik-intel @sanilkumar0 @joshuaranjan @jandres742 @wdamon-intel

jandres742:
this is just change on description, so we could do it on next major version.

jandres742:
targeting v1.8. @wdamon-intel : would that be ok?

wdamon-intel:
@matcabral what do you think?

matcabral:
@aravindksg is your proposal limited only to zes_engine_stats_t or are you looking at doing it for all spec entries for microseconds? I think being able to draw a timeline is a desirable feature.  However, if this is only for zes_engine_stats_t i agree you would also need a way to query the source, to allow drawing a timeline with other timestamps. E.g. metrics.

aravindksg:
> @aravindksg is your proposal limited only to zes_engine_stats_t or are you looking at doing it for all spec entries for microseconds? I think being able to draw a timeline is a desirable feature. However, if this is only for zes_engine_stats_t i agree you would also need a way to query the source, to allow drawing a timeline with other timestamps. E.g. metrics.

@matcabral this is limited to only zes_engine_stats_t. For this usage scenario, user will only be using these values for a calculation of percentage utilization. so they could plot an utilization % graph against, but time I guess would be "host"/OS gettime rather than depending on the value reported in engine_stats_t. The values in engine_stats_t are to reflect HW counters (and not necessarily to infer time for the user)

matcabral:
Ok, I see that the proposal is to only remove it from activeTime and not the timestamp.  I agree with the idea of the change, but leaves few loopholes:

- Activetime: will rename to activecounter? 
- The formula in the description goes with time / time to get a % . will end with a unit of 1/time ?

Bw, also related to the process mapping the timestamps. The description explicitly says not do to it. But following what we did in metrics (synchronize metrics timestamps API), we should consider adding an API allow doing this in sysman.

aravindksg:
> 



[...]

PR #208, opened 2023-09-01
Remove time units for activeTime and timestamp fields

aravindksg:
Addresses #197


aravindksg:
@saik-intel @sanilkumar0 @joshuaranjan @jandres742 @wdamon-intel

scripts/sysman/engine.yml
```diff
@@ -69,11 +69,13 @@ name: $s_engine_stats_t
 members:
     - type: uint64_t
       name: "activeTime"
-      desc: "[out] Monotonic counter for time in microseconds that this resource is actively running workloads."
+      desc: |
+            [out] Monotonic counter where the resource is actively running workloads.
```
jandres742, line 73: what is the driver returning today, us or raw counters?

aravindksg: currently it is us, but data is fundamentally used only to calculate % utilization per formula. so obfuscating units from spec pov allows implementation flexibility to report raw data in case underlying HW is counting at different frequencies than us/ms.

scripts/sysman/engine.yml
```diff
@@ -69,11 +69,13 @@ name: $s_engine_stats_t
...
+            Time units are intentionally obfuscated since the activeTime value is only intended for calculating utilization percentage as noted above.
     - type: uint64_t
       name: "timestamp"
       desc: |
-            [out] Monotonic timestamp counter in microseconds when activeTime counter was sampled.
+            [out] Monotonic counter when activeTime counter was sampled.
             This timestamp should only be used to calculate delta time between snapshots of this structure.
```
joshuaranjan, line 79: Can we remove "time" from here as well.
I think this entire sentence could be removed, since we cannot calculate time using the timestamp (after this change).

aravindksg: you mean from the name of the field? that we can't remove in minor version update since it would break compatibility.

joshuaranjan: > you mean from the name of the field? that we can't remove in minor version update since it would break compatibility.

No in "to calculate delta time betw", we are mentioning that delta time could be calculated.
So either we could have "to calculate delta betw" or may be we could remove the sentence

aravindksg: @joshuaranjan fixed

joshuaranjan:
lgtm

scripts/sysman/engine.yml
jandres742: wouldnt this cause a problem for customers, since they are assuming this is returned in us?

aravindksg: as far as we know customers are not using the fact that it's in us in any way currently. i.e, they just use the formula to calculate % utilization

jandres742: @aravindksg could you share details on that formula? doesn't the formula assume the value to be on us?

aravindksg: formula is from here: https://spec.oneapi.io/level-zero/latest/sysman/api.html#zes-engine-stats-t

```
Percent utilization is calculated by taking two snapshots (s1, s2) and using the equation: 
util = (s2.activeTime - s1.activeTime) / (s2.timestamp - s1.timestamp)
```

so here, doesn't really matter what units we have, we are really interested to just find the ratio. it just matters that `activeTime` and `timestamp` values are relatable so that this calculation works.

saik-intel: @aravindksg  you are removing restriction to represent in us, instead this structure could represent in any time either ms or ns or us right ?  But don;t we need to make sure both active time and time stamp should represent in same time unit whether it is sec or ms or us or ns ?

aravindksg: @saik-intel regarding why we want to remove time units completely, pasting reply from above:

> obfuscating units from spec pov allows implementation flexibility to report raw data in case underlying HW is counting at different frequencies than us/ms

> don;t we need to make sure both active time and time stamp should represent in same

yes, both these values need to be relatable. I can add statement to that effect. we already have following statement in the spec: `Never take the delta of this timestamp with the timestamp from a different structure since they are not guaranteed to have the same base`. so can apply similar strict wording to enforce that activeTime and timestamp are in same units

jandres742: thanks @aravindksg. 

> so obfuscating units from spec pov allows implementation flexibility to report raw data in case underlying HW is counting at different frequencies than us/ms.

could you elaborate on the need for that flexibility?

>formula is from here: https://spec.oneapi.io/level-zero/latest/sysman/api.html#zes-engine-stats-t

right, there's a formula, but the spec doesn't say those fields could be used by the user in other way. here for instance:

> uint64_t activeTime
[out] Monotonic counter for time in microseconds that this resource is actively running workloads.

it doesnt say that the activeTime can only be used on that formula. Furthermore, if we are saying that activeTime and timestamp are only useful for the formula, why dont we just return the value of the formula and have the driver implementation calculating it internally?

aravindksg: > thanks @aravindksg.
> 
> > so obfuscating units from spec pov allows implementation flexibility to report raw data in case underlying HW is counting at different frequencies than us/ms.
> 
> could you elaborate on the need for that flexibility?
> 

one example would be that the wall clock times could be different in VF vs PF domain. So when PF queries engine stats for all VFs, and if it infers in time (since we currently have "microseconds" wording) then it would be unclear to user what reference point of time would be (since environments are different)

> > formula is from here: https://spec.oneapi.io/level-zero/latest/sysman/api.html#zes-engine-stats-t
> 
> right, there's a formula, but the spec doesn't say those fields could be used by the user in other way. here for instance:
> 
> > uint64_t activeTime
> > [out] Monotonic counter for time in microseconds that this resource is actively running workloads.
> 
> it doesnt say that the activeTime can only be used on that formula. Furthermore, if we are saying that activeTime and timestamp are only useful for the formula, why dont we just return the value of the formula and have the driver implementation calculating it internally?

yes, implementation could do it, but the calculation is applied for 2 snapshots taken at two different time intervals. So if driver has to do it, then it might need a "timeout" param so that driver can block for that value and take 2 snapshots of engine activity.

jandres742: > Time units are intentionally obfuscated

to

> Time units are implementation specific

aravindksg: @jandres742 any rationale for using "implementation specific" wording rather than "obfuscated"? because terming "time units" as implementation specific is unclear IMHO.

jandres742: in terms of specifications, "implementation specific" is more used than "obfuscated". Basically what we are saying is that user shouldn't assume anything about the time units.

@wdamon-intel : what do you think?

wdamon-intel:
I think “implementation specific” is better. Obfuscation is not a meaning any spec should convey. Another term that might work OK here is “opaque”, but "implementation specific" is preferable.  It allows for the most flexibility in terms of implementation, while informing the user that the type or representation of some object or metric is purposefully opaque for their purposes.
________________________________

scripts/sysman/engine.yml
aravindksg: fixed

joshuaranjan (approved):

jandres742 (approved):

wdamon-intel (approved):

PR #210, opened 2023-09-06
Add correctly callbacks definitions to headers

jandres742:
Callbacks for functions declared for a specific version were not being added to ze_api.h, forcing us to have them defined in include/layers/zel_tracing_register_cb.h, which is not correct.

So fix it so all are added to ze_api.h. by the time execution reaches the line being fixed here, versioning has already been taken into account, so only the functions for a specific version have been defined.

### DIFF
--- a/scripts/templates/helper.py
+++ b/scripts/templates/helper.py
@@ -988,7 +988,7 @@ Public:
 def get_pfncbtables(specs, meta, namespace, tags):
     tables = []
     for cname in sorted(meta['class'], key=lambda x: meta['class'][x]['ordinal']):
-        objs = get_class_function_objs(specs, cname, 1.0)
+        objs = get_class_function_objs(specs, cname)
         if len(objs) > 0:
             name = get_table_name(namespace, tags, {'class': cname})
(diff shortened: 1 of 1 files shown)


wdamon-intel (dismissed):

scripts/templates/helper.py
```diff
@@ -988,7 +988,7 @@ def get_pfntables(specs, meta, namespace, tags):
 def get_pfncbtables(specs, meta, namespace, tags):
     tables = []
     for cname in sorted(meta['class'], key=lambda x: meta['class'][x]['ordinal']):
-        objs = get_class_function_objs(specs, cname, 1.0)
+        objs = get_class_function_objs(specs, cname)
```
bmyates, line 991: Adding APIs released after 1.0 will change the size and layout of ze_callbacks_t which will be a backwards compatibility break.  That is why I originally limited this to 1.0

jandres742:
Moving to draft, as callbacks are now in the loader, where the official tracing layer is.

wdamon-intel:
Closing as "won't fix".

Issue #209, opened 2023-09-01
Add mechanism to gather memory utilization info for all VFs


aravindksg:
Add extension capability to current API `zesMemoryGetState` to gather memory utilization info for all VFs from PF (similar to `zesGetEngineActivityExt`)

aravindksg:
@sanilkumar0 @jandres742 @joshuaranjan @saik-intel @wdamon-intel

jandres742:
@sanilkumar0 : could you take a look?

sanilkumar0:
Similar to Engine activity reported perf VF, Memory usage reporting per VF is required. Create  zesMemoryGetStateExt with 'free' and 'size' fields

saik-intel:
@aravindksg  is PR raised for this ?

aravindksg:
Closing as we are moving to new set of VF interfaces with #248

PR #216, opened 2023-09-08
Add extension for reporting memory state in sriov environments

aravindksg:
Addresses: #209

### DIFF
--- a/scripts/sysman/EXT_EXP_MemState.rst
+++ b/scripts/sysman/EXT_EXP_MemState.rst
@@ -1,0 +1,78 @@
+<%
+import re
+from templates import helper as th
+%><%
+    OneApi=tags['$OneApi']
+    x=tags['$x']
+    X=x.upper()
+    s=tags['$s']
+    S=s.upper()
+%>
+:orphan:
+
+.. _ZES_extension_memory_state:
+
+==========================
+ Memory State Extension
+==========================
+
+API
+----
+
+* Functions
+
+    * ${s}MemGetStateExp
+
+* Enumerations
+
+    * ${s}_mem_state_exp_version_t
+
+* Structures
+
+    * ${s}_engine_exp_properties_t
+
+Memory State
+~~~~~~~~~~~~
+
+This extension provides capability to user to gather the activity stats for each Virtual Function (VF) associated with memory module.
+Intended use case of the feature is to query utilization stats from Physical Function (PF) interface. However, the extension
+may return error code ${X}_RESULT_ERROR_UNSUPPORTED_FEATURE if an environment does not support it.
+If VF has no memory utilization to report, then implementation shall reflect that appropriately in ${s}_mem_state_t struct so that percentage
+allocation calculation results in value of 0. In this case, percentage free could be 100%.
+If PF has no memory utilization to report, then implementation shall reflect that appropriately in ${s}_mem_state_t struct which is
+at index 0 of the output vector such that allocation calculation results in value of 0. In this case, percentage free could be 100%.
+
+The following pseudo-code demonstrates a sequence for obtaining the memory allocation percentage for all Virtual Functions from Physical Function environment:
+
+.. parsed-literal::
+
+    // Query for memory properties and gather number of VF instances by passing extension structure.
+    // Number of VFs is per memory handle and equates to number of VFs actively using
+    // the resource represented by the memory handle.
+    // Together with pCount from GetStateExp, user can gather the full view of memory state
+    // for all VFs and PF as shown below
+
+    ${s}_mem_properties_t memProps;
+    ${s}_mem_exp_properties_t memExpProps;
+    memProps.stype = ZES_STRUCTURE_TYPE_MEMORY_PROPERTIES;
+    memProps.pNext = &memExpProps;
+    memExpProps.stype = ZES_STRUCTURE_TYPE_MEMORY_EXP_PROPERTIES;
+
+    ${s}MemGetProperties(hMemory, &memProps);
+
+    // Query count of memory states to gather
+    // this count is inclusive of PF and N number of VFs
+    // i.e. if memStateCount = 4, then this implies, one PF state at index 0
+    //      of output vector followed by memory states for 3 VFs
+    uint32_t memStateCount = 0;
+    {s}MemGetStateExp(hMemory, &memStateCount, nullptr);
+
+    // Gather memory states for all VFs and PF.
+    // memory  allocation per formula in ${s}_mem_state_t
+    ${s}_mem_state_t* memStates = (${s}_mem_state_t*) allocate(memStateCount * sizeof(${s}_mem_state_t));
+    {s}MemGetStateExp(hMemory, &memStateCount, memStates);
+
+    for (uint32_t idx = 0; idx < memStateCount; ++idx) {
+        double allocation = 100 * (memStates[idx].size - memStates[idx].free)/(memStates[idx].size;
+        output("%s Allocation of memory module type pertaining to this handle = %.2f\n", (idx == 0) ? "PF" : "VF", allocation);
+    }
--- a/scripts/sysman/memState.yml
+++ b/scripts/sysman/memState.yml
@@ -1,0 +1,71 @@
+#
+# Copyright (C) 2023 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+#
+# See YaML.md for syntax definition
+#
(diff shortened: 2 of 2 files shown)


aravindksg:
@sanilkumar0 @jandres742 @joshuaranjan @saik-intel @wdamon-intel @jitendrasharma1989

scripts/sysman/memState.yml
```diff
@@ -0,0 +1,71 @@
...
+            This count represents the number of VF instances that are actively using the resource represented by the memory module handle.
+details:
+    - "This structure may be passed to $sMemoryGetProperties by having the pNext member of $s_mem_properties_t point at this struct."
+    - "Used for SRIOV per VF device utilization by $s_mem_type_t"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Get memory state for Physical Function (PF) and each Virtual Function (VF) associated with memory module."
+version: "1.8"
+class: $sMemory
+name: GetStateExp
```
jandres742, line 48: can we name this less generic? this is state of Physical and Virtual functions, so maybe something like zesMemoryGetVirtualAndPhyscalFunctionsStateExp, but maybe that's too long.

aravindksg: no, this would be zesMemoryGetStateExp, so intention is to gather memory state (alloc/free counters) for VF and PF

scripts/sysman/EXT_EXP_MemState.rst
```diff
@@ -0,0 +1,78 @@
...
+%><%
+    OneApi=tags['$OneApi']
+    x=tags['$x']
+    X=x.upper()
+    s=tags['$s']
+    S=s.upper()
+%>
+:orphan:
+
+.. _ZES_extension_memory_state:
```
joshuaranjan, line 13: According to the SAS 
"An Intel SW agent (XPUM/L0/KMD?) running in the VM will collect graphics memory consumption statistics from the OS video memory manager. This agent will 'push' the data down the stack and to the PF KMD via the VF2PF relay mechanism. "

According to https://docs.intel.com/documents/GfxSWAT/Common/GFX_SRIOV_SAS/ATS-M_PVC/Graphics%20SRIOV%20SAS/GraphicsSRIOVSAS.html#pf-kmd-sriov-startup-runtime , Telemetry From in-VM GPU SW Sources section , we need a Sysman API to push the memory utilization data down to the Guc

Please share whether this API will be covered in this Spec.


Thanks

scripts/sysman/memState.yml
```diff
@@ -0,0 +1,71 @@
...
+    - "The application may call this function from simultaneous threads."
+    - "The implementation of this function should be lock-free."
+params:
+    - type: $s_mem_handle_t
+      name: hMemory
+      desc: "[in] Handle for the component."
+    - type: uint32_t*
+      name: pCount
+      desc: |
+            [in,out] Pointer to the number of VF memory state structures.
```
joshuaranjan, line 59: VF + PF ?

scripts/sysman/EXT_EXP_MemState.rst
```diff
@@ -0,0 +1,78 @@
...
+    * ${s}_engine_exp_properties_t
+
+Memory State
+~~~~~~~~~~~~
+
+This extension provides capability to user to gather the activity stats for each Virtual Function (VF) associated with memory module.
+Intended use case of the feature is to query utilization stats from Physical Function (PF) interface. However, the extension
+may return error code ${X}_RESULT_ERROR_UNSUPPORTED_FEATURE if an environment does not support it.
+If VF has no memory utilization to report, then implementation shall reflect that appropriately in ${s}_mem_state_t struct so that percentage
+allocation calculation results in value of 0. In this case, percentage free could be 100%.
```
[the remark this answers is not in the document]
aravindksg: right, but if  `size == 0`, then user needs to make sure not to perform a divide by 0. so may need to mention that in pseudo-code. @sanilkumar0 could you also clarify this behavior? i.e, if we want to report no memory utilization for a VF, then guidance is to set 0 for size?

scripts/sysman/EXT_EXP_MemState.rst
```diff
@@ -0,0 +1,78 @@
...
+    // Number of VFs is per memory handle and equates to number of VFs actively using
+    // the resource represented by the memory handle.
+    // Together with pCount from GetStateExp, user can gather the full view of memory state
+    // for all VFs and PF as shown below
+
+    ${s}_mem_properties_t memProps;
+    ${s}_mem_exp_properties_t memExpProps;
+    memProps.stype = ZES_STRUCTURE_TYPE_MEMORY_PROPERTIES;
+    memProps.pNext = &memExpProps;
+    memExpProps.stype = ZES_STRUCTURE_TYPE_MEMORY_EXP_PROPERTIES;
```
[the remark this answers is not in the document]
aravindksg: sure, I'll need to rename API as well for this anyway so shall update

scripts/sysman/memState.yml
aravindksg: yes, will reformat this desc

scripts/sysman/EXT_EXP_MemState.rst
```diff
@@ -0,0 +1,78 @@
...
+==========================
+ Memory State Extension
+==========================
+
+API
+----
+
+* Functions
+
+    * ${s}MemGetStateExp
```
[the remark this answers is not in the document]
saik-intel: zesMemGetVirtualizedStateExt would be better instead of Exp, not sure why Experimental you want ?

scripts/sysman/EXT_EXP_MemState.rst
```diff
@@ -0,0 +1,78 @@
...
+
+* Enumerations
+
+    * ${s}_mem_state_exp_version_t
+
+* Structures
+
+    * ${s}_engine_exp_properties_t
+
```
saik-intel, line 33: same here

scripts/sysman/EXT_EXP_MemState.rst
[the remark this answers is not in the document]
aravindksg: given that we have had to make corrections in sysman API recently, preference was to go with experimental extension first. If that works, then we can always promote to standard extension or standard interfaces

scripts/sysman/EXT_EXP_MemState.rst
```diff
@@ -0,0 +1,78 @@
...
+----
+
+* Functions
+
+    * ${s}MemGetStateExp
+
+* Enumerations
+
+    * ${s}_mem_state_exp_version_t
+
```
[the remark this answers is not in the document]
aravindksg: exp for now

scripts/sysman/EXT_EXP_MemState.rst
aravindksg: exp for now

scripts/sysman/memState.yml
jandres742: so does that mean that [zesMemoryGetState](https://spec.oneapi.io/level-zero/latest/sysman/api.html?highlight=zesmemorygetstate#_CPPv417zesMemoryGetState16zes_mem_handle_tP15zes_mem_state_t) would be deprecated? if it is for all memory, do we need to mention about VF and PF?

jitendrasharma1989: @aravindksg Couldnt we extend existing zesMemroyGetState by adding a new field in zes_structure_type_t. As i feel that current name of API zesMemoryGetState is good. All other names sounds too much elaborative.

aravindksg: @jandres742 no `zesMemoryGetState` is still valid and might return valid info for environments when info can be retrieved. But we need separate API for gathering virtualized memory state (similar to how we do for engine stats via `zesEngineGetActivityExt`) from PF/VF domains. hence the proposal here. 

@jitendrasharma1989 each structure only has one struct type. don't think we can define new struct type and use with existing structure. 

As mentioned below, thinking of going with `zesMemGetVirtualizedStateExp` (which is probably not too long)

scripts/sysman/EXT_EXP_MemState.rst
aravindksg: @joshuaranjan closed on this offline with @sanilkumar0 . looks like we don't need API for that

aravindksg:
Closing this PR as we are moving to new set of VF interfaces with #248

Issue #102, opened 2023-03-22
IPC event pool device list must match event pool device list


zzdanowicz:
1. Normal event pool can get device list from the context or it might be a custom list of devices provided in API function arguments.
2. When creating IPC event pool from IPC handle, device list is taken from the context.
3. Device lists of normal event pool and IPC event pool must match, otherwise driver will run into problems. 
4. Spec should explicitly prohibit it.

jandres742:
@zzdanowicz : should match or could it be a sub-set? Let's say that in a 4 accelerator system, the even pool is created with acc0, acc1, and acc2. Then when opening the event pool handle, we do it with a context that only has:

1.
acc1 and acc3: could the driver implicitly detect that only acc1 can access the event pool?

2.
acc1 and acc2: could the driver implicitly detect that acc0 won't have access, given doesnt belong to the context?

Maybe we could add something like this to  zeEventPoolOpenIpcHandle:

"After creation, the event pool is accessible only for the devices with which the pool was originally created that also belong to the context passed to zeEventPoolOpenIpcHandle"

zzdanowicz:
I think they should match.  And there is very little, almost to no maneuver:
2. If the device is not in event pool device list, events from that pool cannot be used on this device. 
3. Opening IPC event pool with device list (taken from the context used to open IPC event pool), means all devices from that context are able to use events from that IPC event pool.

So if IPC context has different device list than the device list of remote event pool, then you have clear contradiction between point 1 and 2. 

4. But there are practical problems, even with the "acc1" and "acc3" scenario. In order to pass that information between processes you could quickly expend all available size of IPC opaque handle (and would just kill definition of memory layout of that opaque handle, as device lists can be variable and even include sub-devices. 
5. That stands also against L0 API explicit and precise use - all implicit work adds complexity (tracking which device can or cannot use event) at least should be considered as not needed.
6. Definition of device list is also connected to the event pool allocation directly coupled with the allocation being visible and available on that devices.
7. Most important practical reason is the memory layout (number of packets) of the single event from the event pool is connected with all devices. And using such event right now will lead to memory overruns or invalid event synchronization.

Enforcing the use the exact device list for both event pools makes them behave exactly the same without any worry, implicit works and tracking. And it will save a lot of problem in case of that unregulated use.

Adding @MichalMrozek

jandres742:
thanks @zzdanowicz . Those are good points to have the requirement that the list of devices in the importer and exporter. Thanks.

[...]

PR #219, opened 2023-09-18
Clarify usage of IPC event pools

jandres742:
Resolves: #102

### DIFF
--- a/scripts/core/event.yml
+++ b/scripts/core/event.yml
@@ -227,6 +227,13 @@ decl: static
 details:
     - "Multiple calls to this function with the same IPC handle will return unique event pool handles."
     - "The event handle in this process should not be freed with $xEventPoolDestroy, but rather with $xEventPoolCloseIpcHandle."
+    - |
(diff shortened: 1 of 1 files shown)


MichalMrozek (commented):
The requirement is other way around.
We can still import event pool created with root device on sub device. (so there is no requirement that on importer side we need to have at least x sub devices)
But we cannot import event pool created with sub device on root device containing 2 sub devices.

jandres742:
> The requirement is other way around. We can still import event pool created with root device on sub device. (so there is no requirement that on importer side we need to have at least x sub devices) But we cannot import event pool created with sub device on root device containing 2 sub devices.

thanks @MichalMrozek . Please see update.

MichalMrozek (commented):
lgtm

jandres742:
thanks @MichalMrozek .

@wdamon-intel : please check.

wdamon-intel (approved):

MichalMrozek (approved):

Issue #145, opened 2023-06-07
New in-order Event type


BartoszDunajski:
As continuation of discussion about #81, we decided that new Event type for in-order command list mode is required.

Requirements (open to discuss):
- In-order CL design assumes internal storage to track implicit dependencies. This is incremental counter
- Event inherits that value from CL during append API call
- New Event flag is required to create such Event
- New type can be signaled only from in-order CL
- Reset of that Event is forbidden
- Should we disallow host signaling too? 
- Timestamp mode is still allowed
- Event can be reused multiple times without reset. New API call overrides previous state (counter to wait)
- Wait call (GPU/CPU) on reused Event waits only for latest saved counter value
- Event can be reused on different command lists. Wait call (GPU/CPU) waits only for latest used CL and assigned counter value

MichalMrozek:
Countered events doesn't make sense for out of order command lists, so I agree requirement should be to have it only for in order command lists.

"
Should we disallow host signaling too?
"

Yes host signaling would not possible and must be disallowed.

One more point, we need to clearly define what happens for non immediate command lists.
One option is to disallow those events in those scenario.
Another option is to allow them, but with clear semantic how they can be reused and how latests state is reflected, as it will not happen on append anymore due to recording.

MichalMrozek:
We would need additional flag that would specify that countered events can be used on recorded command lists.
There will be additional handling needed in such scenarios and we do not want applications to pay it if they only want to use those countered events on immediate command lists.

jandres742:
@MichalMrozek , @BartoszDunajski : please check https://github.com/oneapi-src/level-zero-spec/pull/221

BartoszDunajski:
> We would need additional flag that would specify that countered events can be used on recorded command lists. There will be additional handling needed in such scenarios and we do not want applications to pay it if they only want to use those countered events on immediate command lists.

Should we make those flags mutually exclusive? For example event generated on non-immediate cmd lists cannot be used on immediate?

MichalMrozek:
Right now user can represent if it wants:
- event for immediate command list 
- event for non immedaite command lists
- event for both immediate and non immediate command list

So I don't think we need to make them mutually exclusive.

PR #221, opened 2023-09-19
Add experimental extension for counter-based events

jandres742:
Resolves: #145

### DIFF
--- a/scripts/core/EXT_Exp_CounterBasedEvents.rst
+++ b/scripts/core/EXT_Exp_CounterBasedEvents.rst
@@ -1,0 +1,49 @@
+<%
+import re
+from templates import helper as th
+%><%
+    OneApi=tags['$OneApi']
+    x=tags['$x']
+    X=x.upper()
+%>
+:orphan:
+
+.. _ZE_experimental_event_counter_based:
+
+=========================
+ Counter-Based Events Extension
+=========================
+
+API
+----
+ 
+* Enumerations
+
+    * ${x}_event_counter_based_exp_flags_t
+    * ${x}_event_counter_based_exp_version_t
+
+* Structures
+
+    * ${x}_event_counter_based_exp_desc_t
+
+Counter-Based Events
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, events in Level Zero contains one of two states: signaled or not signaled.
+Signaling of an event, as well as resetting its state, can be done from either host or
+device.
+
+Another way of tracking the state of an event is by using a counter, which is incremented
+every time a task has completed in the device. Using a counter-based event may be
+provide Level Zero driver implementations with the opportunity for both
+functional and performance optimizations.
+
+The following recommendations and restrictions apply to counter-based events:
+
+- Counter-based events can be used only with in-orders lists, whether those are regular or immediate.
+- Counter-based events must not be reset, i.e., a call to ${x}EventHostSynchronize or ${x}CommandListAppendEventReset
+  is not allowed. Counter-based events may be reused multiple times without a need for reset.
+- Counter-based events must not be signaled from host, i.e., a call to ${x}EventHostSignal is not allowed.
+- Counter-based events may be used on multiple command lists.
+- Querying a counter-based event returns only the last saved counter value from the last command list that incremented it.
+- Synchronizing on a counter-based event waits only the last saved counter value from the last command list that incremented it.
--- a/scripts/core/counterbasedevent.yml
+++ b/scripts/core/counterbasedevent.yml
@@ -1,0 +1,51 @@
+#
+# Copyright (C) 2019-2022 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+#
+# See YaML.md for syntax definition
+#
+--- #--------------------------------------------------------------------------
+type: header
+desc: "Intel $OneApi Level-Zero Extension APIs for Counter-based Events"
+version: "1.8"
+--- #--------------------------------------------------------------------------
+type: macro
+desc: "Counter-based Events Extension Name"
+version: "1.4"
+name: $X_EVENT_COUNTER_BASED_EXP_NAME
+value: '"$X_experimental_event_counter_based"'
+--- #--------------------------------------------------------------------------
+type: enum
+desc: "Counter-based Events Extension Version(s)"
+version: "1.8"
+name: $x_event_counter_based_exp_version_t
+etors:
+    - name: "1_0"
+      value: "$X_MAKE_VERSION( 1, 0 )"
+      desc: "version 1.0"
+--- #--------------------------------------------------------------------------
+type: enum
+desc: "Supported event flags for defining counter-based events. Must "
+class: $xEvent
+name: $x_event_counter_based_exp_flags_t
+etors:
+    - name: IMMEDIATE
+      desc: "Counter-based event is used for immediate command lists (default)"
+    - name: NON_REGULAR
+      desc: "Counter-based event is used for non-immediate command lists"
+--- #--------------------------------------------------------------------------
+type: struct
+desc: "Event descriptor for counter-based events. This structure may be passed to $xEventCreate as pNext member of $x_event_desc_t."
+class: $xEvent
+version: "1.8"
+name: $x_event_counter_based_exp_desc_t
+base: $x_base_desc_t
+members:
+    - type: $x_event_counter_based_exp_flags_t
+      name: flags
+      desc: |
+            [in] mode flags.
+            must be 0 (default) or a valid value of $x_event_counter_based_exp_flag_t
+            default behavior is counter-based event is only used for immediate command lists.
+      init: "0"
--- a/scripts/core/common.yml
+++ b/scripts/core/common.yml
@@ -508,6 +508,10 @@ etors:
       desc: $x_rtas_geometry_aabbs_exp_cb_params_t
       version: "1.7"
       value: "0x00020013"
+    - name: COUNTER_BASED_EVENT_EXP_DESC
+      desc: $x_event_counter_based_exp_desc_t
+      version: "1.8"
+      value: "0x00020014"
     - name: COMMAND_GRAPH_EXP_DESC
       desc: $x_command_graph_exp_desc_t
       version: "2.0"


jandres742:
this is how rendered would look like:

image

BartoszDunajski:
I think you should add more details about usage restrictions

scripts/core/counterbasedevent.yml
```diff
@@ -0,0 +1,51 @@
...
+      desc: "version 1.0"
+--- #--------------------------------------------------------------------------
+type: enum
+desc: "Supported event flags for defining counter-based events. Must "
+class: $xEvent
+name: $x_event_counter_based_exp_flags_t
+etors:
+    - name: IMMEDIATE
+      desc: "Counter-based event is used for immediate command lists (default)"
+    - name: NON_REGULAR
```
MichalMrozek, line 35: do you mean NON_IMMEDIATE here?

jandres742: thanks @MichalMrozek . Fixed.

jandres742:
@MichalMrozek , @BartoszDunajski : please take a look at latest changes.

scripts/core/counterbasedevent.yml
aravindksg: do we need two flags? say if we just have IMMEDIATE, then the absence of that flag would mean REGULAR cmdlists?

jandres742: thanks @aravindksg . I was thinking on that. However, it is better to be explicit. in any case, as you see, we are saying that default value is zero, which is IMMEDAITE, so the user only needs to pass a flag (NON_REGULAR) if they want for regular. 

```cpp
type: $x_event_counter_based_exp_flags_t
      name: flags
      desc: |
            [in] mode flags.
            must be 0 (default)
```

scripts/core/EXT_Exp_CounterBasedEvents.rst
```diff
@@ -0,0 +1,49 @@
...
+Counter-Based Events
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, events in Level Zero contains one of two states: signaled or not signaled.
+Signaling of an event, as well as resetting its state, can be done from either host or
+device.
+
+Another way of tracking the state of an event is by using a counter, which is incremented
+every time a task has completed in the device. Using a counter-based event may be
+provide Level Zero driver implementations with the opportunity for both
```
aravindksg: statement reformatting: " may be provide" to "may provide"

scripts/core/EXT_Exp_CounterBasedEvents.rst
```diff
@@ -0,0 +1,49 @@
...
+functional and performance optimizations.
+
+The following recommendations and restrictions apply to counter-based events:
+
+- Counter-based events can be used only with in-orders lists, whether those are regular or immediate.
+- Counter-based events must not be reset, i.e., a call to ${x}EventHostSynchronize or ${x}CommandListAppendEventReset
```
aravindksg: statement reformatting: "counter-based events must not be reset or synchronized. i.e., ..."
also- zeEventHostReset would not be allowed right?

scripts/core/EXT_Exp_CounterBasedEvents.rst
```diff
@@ -0,0 +1,49 @@
...
+  is not allowed. Counter-based events may be reused multiple times without a need for reset.
+- Counter-based events must not be signaled from host, i.e., a call to ${x}EventHostSignal is not allowed.
+- Counter-based events may be used on multiple command lists.
```
aravindksg, line 48: do we want to clarify further? i.e, for use with multiple command lists, users must create different events.

scripts/core/EXT_Exp_CounterBasedEvents.rst
```diff
@@ -0,0 +1,49 @@
...
+- Querying a counter-based event returns only the last saved counter value from the last command list that incremented it.
+- Synchronizing on a counter-based event waits only the last saved counter value from the last command list that incremented it.
```
aravindksg, line 50: we are saying above that `EventHostSynchronize` is not allowed but here you have line stating behavior for syncing on the event.

scripts/core/counterbasedevent.yml
```diff
@@ -0,0 +1,51 @@
...
+      desc: "Counter-based event is used for non-immediate command lists"
+--- #--------------------------------------------------------------------------
+type: struct
+desc: "Event descriptor for counter-based events. This structure may be passed to $xEventCreate as pNext member of $x_event_desc_t."
```
aravindksg, line 39: do we want to consider adding flag for event pools for counter based events? i.e, would there be value in creating an event pool where events from that pool are only used as counting events?

scripts/core/EXT_Exp_CounterBasedEvents.rst
aravindksg: apologies, please ignore this, i see you addressed it in follow up commit

scripts/core/EXT_Exp_CounterBasedEvents.rst
```diff
@@ -0,0 +1,50 @@
...
+every time a task has completed in the device. Using a counter-based event may provide
+Level Zero driver implementations with the opportunity for both functional and performance
+optimizations.
+
+The following recommendations and restrictions apply to counter-based events:
+
+- Counter-based events can be used only with in-orders lists, whether those are regular or immediate.
+- Counter-based events must not be reset, i.e., a call to ${x}EventHostReset or ${x}CommandListAppendEventReset
+  is not allowed.
+- Counter-based events may be reused multiple times without a need for reset.
```
MichalMrozek, line 46: This may be confusing without first explaining how they work.
Can you add a sentence that they automatically track the last signaling usage and that passing the event to new call that will signal it, automatically updates what event represents?
I would also add that passing event to be signaled on new calls, doesn't require the host or device to wait until previous signaling / dependency resolution is completed.

scripts/core/EXT_Exp_CounterBasedEvents.rst
MichalMrozek: why multiple events are needed?
you can have one event and create in order sequence of execution spanning across multiple command lists.

scripts/core/counterbasedevent.yml
MichalMrozek: Yes we need flag for event memory pool as those events doesn't require pool memory to be allocated for them, the storage is inherited from command list.

scripts/core/EXT_Exp_CounterBasedEvents.rst
aravindksg: apologies, misread the original statement, this looks good

scripts/core/EXT_Exp_CounterBasedEvents.rst
jandres742: thanks, changed.

scripts/core/EXT_Exp_CounterBasedEvents.rst
jandres742: thanks. added.

scripts/core/counterbasedevent.yml
jandres742: @MichalMrozek : sorry, could you clarify the comment about event pool? do we want the counter-based flags to be part of the zeEventPoolCreate or zeEventCreate, or both?

scripts/core/EXT_Exp_CounterBasedEvents.rst
jandres742: @MichalMrozek : thanks. please see proposal below:

```md
- Counter-based events can be used only with in-orders lists, whether those are regular or immediate.
- Counter-based events must not be reset, i.e., a call to ${x}EventHostReset or ${x}CommandListAppendEventReset
  is not allowed.
- Counter-based events must not be signaled from host, i.e., a call to ${x}EventHostSignal is not allowed.
- Counter-based events may be reused multiple times without a need for reset.
- Counter-based events may be used on multiple command lists.
- Querying a counter-based event queries only the last saved counter value from the last command list that
  incremented it, i.e., a signaled counter-based event always represents the completion of the last call to which
  it was passed as signal event.
- Synchronizing on a counter-based event waits only the last saved counter value from the last command list
  that incremented it.
- A counter-based event may be passed as signaling event for a new append call without needing to wait for
  the signaling of the last call where it was used.
```

scripts/core/counterbasedevent.yml
MichalMrozek: Counter Based Events doesn't need memory backing if they are for immediate command lists, so this must be event pool flag as this is steered to control where allocation happens.
I do not see a need for event flag.
So only change in zeEventPoolCreate flags.

aravindksg: seems to me that we are talking about "counter events" being associated at a command queue level. so would it be better to describe a different struct for this entirely so we can cleanly associate at command queue/command list level rather than shoe-horning it into existing event(pool) struct/desc?

we have fences described here as being associated at command queue level, so something similar for "counter events"?
https://spec.oneapi.io/level-zero/latest/core/PROG.html#fences

MichalMrozek: We want to leverage events, otherwise all append API's would need to be rewritten.

scripts/core/EXT_Exp_CounterBasedEvents.rst
MichalMrozek: ok

scripts/core/counterbasedevent.yml
jandres742: thanks @MichalMrozek .  I have moved the flags to the event pool logic. Please check.

jandres742:
@MichalMrozek , @aravindksg, @BartoszDunajski  : please see latest changes.

scripts/core/EXT_Exp_CounterBasedEvents.rst
```diff
@@ -0,0 +1,56 @@
...
+----
+ 
+* Enumerations
+
+    * ${x}_event_pool_counter_based_exp_flags_t
+    * ${x}_event_pool_counter_based_exp_version_t
+
+* Structures
+
+    * ${x}_event_pool_counter_based_exp_desc_t
```
aravindksg, line 27: is there a CounterBasedEvents.yml file where you have expanded on this descriptor? not seeing that as part of this PR (or am I not looking at right view!?)

jandres742: @aravindksg : files were not added. Please see now.

scripts/core/EXT_Exp_CounterBasedEventPools.rst
```diff
@@ -0,0 +1,56 @@
...
+%><%
+    OneApi=tags['$OneApi']
+    x=tags['$x']
+    X=x.upper()
+%>
+:orphan:
+
+.. _ZE_experimental_event_pool_counter_based:
+
+=========================
```
aravindksg, line 13: minor:  `===` wrapper around the header needs to be as long as the length of the header itself? or doesn't matter? just curious if build complains due to this

scripts/core/counterbasedeventpool.yml
```diff
@@ -0,0 +1,51 @@
...
+# See YaML.md for syntax definition
+#
+--- #--------------------------------------------------------------------------
+type: header
+desc: "Intel $OneApi Level-Zero Extension APIs for Counter-based Event Pools"
+version: "1.8"
+--- #--------------------------------------------------------------------------
+type: macro
+desc: "Counter-based Event Pools Extension Name"
+version: "1.4"
```
aravindksg, line 15: version 1.8

scripts/core/counterbasedeventpool.yml
```diff
@@ -0,0 +1,51 @@
...
+version: "1.8"
+name: $x_event_pool_counter_based_exp_version_t
+etors:
+    - name: "1_0"
+      value: "$X_MAKE_VERSION( 1, 0 )"
+      desc: "version 1.0"
+--- #--------------------------------------------------------------------------
+type: enum
+desc: "Supported event flags for defining counter-based event pools."
+class: $xEventPool
```
aravindksg, line 30: missing version line?

scripts/core/counterbasedeventpool.yml
```diff
@@ -0,0 +1,51 @@
...
+name: $x_event_pool_counter_based_exp_desc_t
+base: $x_base_desc_t
+members:
+    - type: $x_event_pool_counter_based_exp_flags_t
+      name: flags
+      desc: |
+            [in] mode flags.
+            must be 0 (default) or a valid value of $x_event_pool_counter_based_exp_flag_t
+            default behavior is counter-based event pool is only used for immediate command lists.
+      init: "0"
```
aravindksg, line 51: do we need member for `count` which is num events in the pool? we already have a count in `event_pool_desc_t` so are we going to just use that? just making sure so that we document this in a pseudo-code clearly

scripts/core/EXT_Exp_CounterBasedEvents.rst
```diff
@@ -0,0 +1,56 @@
...
+- Counter-based events must not be signaled from host, i.e., a call to ${x}EventHostSignal is not allowed.
+- Counter-based events may be reused multiple times without a need for reset.
+- Counter-based events may be used on multiple command lists.
+- Querying a counter-based event queries only the last saved counter value from the last command list that incremented it,
+  i.e., a signaled counter-based event always represents the completion of the last call to which it was passed as signal event.
+- Synchronizing on a counter-based event waits only the last saved counter value from the last command list that incremented it.
+- A counter-based event may be passed as signaling event for a new append call without needing to wait for the signaling of
+  the last call where it was used.
+
+Counter-based events can be created by passing ${x}_event_pool_counter_based_exp_desc_t to ${x}EventPoolCreate
```
aravindksg, line 55: could you add pseudo-code at the end of this section to show how user can set the `exp_desc_t` and then call `eventPoolCreate` ?

scripts/core/EXT_Exp_CounterBasedEvents.rst
```diff
@@ -0,0 +1,56 @@
...
+- Counter-based events must not be reset, i.e., a call to ${x}EventHostReset or ${x}CommandListAppendEventReset
+  is not allowed.
+- Counter-based events must not be signaled from host, i.e., a call to ${x}EventHostSignal is not allowed.
+- Counter-based events may be reused multiple times without a need for reset.
+- Counter-based events may be used on multiple command lists.
+- Querying a counter-based event queries only the last saved counter value from the last command list that incremented it,
+  i.e., a signaled counter-based event always represents the completion of the last call to which it was passed as signal event.
+- Synchronizing on a counter-based event waits only the last saved counter value from the last command list that incremented it.
+- A counter-based event may be passed as signaling event for a new append call without needing to wait for the signaling of
+  the last call where it was used.
```
aravindksg, line 53: do we want to add notes on effect of "Timestamp" flags with counter events? I think timestamp flags (even if set) would be ignored right? or we want to forbid user from setting it and return error?

also- do we expect IPC events to continue to work with counter events?

scripts/core/EXT_Exp_CounterBasedEventPools.rst
jandres742: fixed in latest

scripts/core/EXT_Exp_CounterBasedEvents.rst
jandres742: timestamps and IPC are orthogonal features, I dont think they cannot work, However, if that's the case, we can add note in the documentation in future follow ups.

scripts/core/EXT_Exp_CounterBasedEvents.rst
jandres742: added

scripts/core/counterbasedeventpool.yml
jandres742: fixed

scripts/core/counterbasedeventpool.yml
jandres742: count still comes from vent_pool_desc_t. here we are only adding a flag.

aravindksg (approved):

wdamon-intel (approved):

BartoszDunajski:
In general looks ok. But usage on Regular CmdLists is not well described here. This is something that may be confusing in the future and its worth to explain it.

Counter assigned to Event is updated on Each "execute" call.

For example:
```
CL1->appendLaunchKernel();
CL1->appendLaunchKernel(signal: Event1);
CL1->appendLaunchKernel();
CL1->close();

CmdQ->execute(CL1); // counter {1 -> 2 -> 3}
CmdQ->execute(CL1); // counter {4 -> 5 -> 6}
```

```
CL2-> appendLaunchKernel(wait: Event1); 
CL2->close();
CmdQ->execute(CL2); // wait for counter == 5
```

CLs executed again:
```
CmdQ->execute(CL1); // counter {7 -> 8 -> 9}
```

```
CmdQ->execute(CL2); // wait for counter == 8
```

jandres742:
> In general looks ok. But usage on Regular CmdLists is not well described here. This is something that may be confusing in the future and its worth to explain it.
> 
> Counter assigned to Event is updated on Each "execute" call.
> 
> For example:
> 
> ```
> CL1->appendLaunchKernel();
> CL1->appendLaunchKernel(signal: Event1);
> CL1->appendLaunchKernel();
> CL1->close();
> 
> CmdQ->execute(CL1); // counter {1 -> 2 -> 3}
> CmdQ->execute(CL1); // counter {4 -> 5 -> 6}
> ```
> 
> ```
> CL2-> appendLaunchKernel(wait: Event1); 
> CL2->close();
> CmdQ->execute(CL2); // wait for counter == 5
> ```
> 
> CLs executed again:
> 
> ```
> CmdQ->execute(CL1); // counter {7 -> 8 -> 9}
> ```
> 
> ```
> CmdQ->execute(CL2); // wait for counter == 8
> ```

thanks @BartoszDunajski . Let me add that clarification in a follow-up patch.

Issue #223, opened 2023-09-19
Add new GPU power domain


aravindksg:
Add ZES_POWER_DOMAIN_GPU to  zes_power_domain_t

wdamon-intel:
Resolved by #224

PR #224, opened 2023-09-19
Add new GPU power domain

aravindksg:
CARD domain is technically for whole PCI card which might include more than just the GPU IP. So, adding separate GPU domain helps to provide GPU specific power related info.

Addresses: #223


aravindksg:
@sanilkumar0 @jandres742 @joshuaranjan @saik-intel @wdamon-intel @jitendrasharma1989

jandres742 (approved):

Issue #218, opened 2023-09-13
Deprecate zesSchedulerSetComputeUnitDebugMode API


aravindksg:
- Given this is debug related, it does not really belong to Scheduler API set
- there is no use case currently for per-device switches for debugger
- For debugger enablement, global envvar is currently used ZET_ENABLE_PROGRAM_DEBUGGING and for device level masking ZE_AFFINITY_MASK is used already

jandres742:
@aravindksg : is this implemented in the L0 GPU driver?

bmyates:
This is implemented in the intel gpu driver for linux. It is not implemented for windows. 

The current implementation is toggling KMD eudebug feature enable by sysfs write. There are no users of this API.  Instead, we instruct users to directly write the sysfs file node. 

ZET_ENABLE_PROGRAM_DEBUGGING is used as UMD enable, and does not directly enable debugging in KMD.

PR #228, opened 2023-09-22
Deprecating compute unit debug mode

aravindksg:
Addresses: #218

### DIFF
--- a/scripts/sysman/scheduler.yml
+++ b/scripts/sysman/scheduler.yml
@@ -22,7 +22,9 @@ etors:
     - name: EXCLUSIVE
       desc: "Any application or context can run indefinitely on the hardware without being preempted or terminated. All pending work for other contexts must wait until the running context completes with no further submitted work."
     - name: COMPUTE_UNIT_DEBUG
-      desc: "This is a special mode that must ben enabled when debugging an application that uses this device e.g. using the Level0 Debug API. It has the effect of disabling any timeouts on workload execution time and will change workload scheduling to ensure debug accuracy."
+      desc:
+        "1.0": "This is a special mode that must ben enabled when debugging an application that uses this device e.g. using the Level0 Debug API. It has the effect of disabling any timeouts on workload execution time and will change workload scheduling to ensure debug accuracy."
+        "1.8": "[DEPRECATED] No longer supported."
 --- #--------------------------------------------------------------------------
 type: struct
 desc: "Properties related to scheduler component"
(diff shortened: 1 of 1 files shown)


aravindksg:
@sanilkumar0 @jandres742 @joshuaranjan @saik-intel @wdamon-intel @jitendrasharma1989 @bmyates

jandres742 (commented):
@bmyates: please +1 from debugger side.

scripts/sysman/scheduler.yml
```diff
@@ -22,7 +22,9 @@ etors:
     - name: EXCLUSIVE
       desc: "Any application or context can run indefinitely on the hardware without being preempted or terminated. All pending work for other contexts must wait until the running context completes with no further submitted work."
     - name: COMPUTE_UNIT_DEBUG
-      desc: "This is a special mode that must ben enabled when debugging an application that uses this device e.g. using the Level0 Debug API. It has the effect of disabling any timeouts on workload execution time and will change workload scheduling to ensure debug accuracy."
+      desc:
+        "1.7": "This is a special mode that must ben enabled when debugging an application that uses this device e.g. using the Level0 Debug API. It has the effect of disabling any timeouts on workload execution time and will change workload scheduling to ensure debug accuracy."
```
jandres742, line 26: No need to version this description. We can keep it. Only needed is the 1.8 description with the deprecation notice.

If we want to version the old description, it should be 1.0, since this enum is valid since then, but again, it is not needed.

aravindksg: so far while deprecating other enums we have added last version up until it is valid. so not sure i follow why it should be "1.0" ?

jandres742: @aravindksg : by versioning with 1.7, you are saying that this text:

"1.7": "This is a special mode that must ben enabled when debugging an application that uses this device e.g. using the Level0 Debug API. It has the effect of disabling any timeouts on workload execution time and will change workload scheduling to ensure debug accuracy."

will appear only with starting at 1.7 version. That means that if we try to build the spec for 1.6 or previous, that text won't appear, and it should.


if we have been doing this that you said:
> so far while deprecating other enums we have added last version up until it is valid

then we have been doing it wrong.

you can confirm this by building the spec for 1.6 with your changes, and confirming whether the text above appears or not.

aravindksg: ah ok I see what you mean. on a different PR I think we deprecated statement that was added in 1.7 . fixing this now

aravindksg: fixed

jandres742 (approved):

Issue #211, opened 2023-09-06
Add mechanism to set and get Secure Version Number for Firmware


aravindksg:
To prevent flashing of older Manufacturing IFWIs that have certain security vulnerabilities or exposures , an anti-rollback scheme needs to be involved. add mechanism to gather min secure version number (SVN) and set SVN number prior to FW flashing

aravindksg:
@sanilkumar0 @jandres742 @wdamon-intel @saik-intel @joshuaranjan @jitendrasharma1989

PR #256, opened 2024-01-09
Add experimental extension to access firmware security version (#211)

sanilkumar0:
Resolves: oneapi-src/level-zero-spec#211


scripts/sysman/EXT_Exp_FirmwareSecurityVersion.rst
```diff
@@ -0,0 +1,30 @@
...
+import re
+from templates import helper as th
+%><%
+    OneApi=tags['$OneApi']
+    s=tags['$s']
+    S=s.upper()
+%>
+:orphan:
+
+.. _ZES_extension_ras_state:
```
aravindksg, line 11: extension name to be changed

scripts/sysman/EXT_Exp_FirmwareSecurityVersion.rst
```diff
@@ -0,0 +1,30 @@
...
+
+======================================
+ Firmware security firmware version
```
aravindksg, line 14: firmware word is repeated

scripts/sysman/EXT_Exp_FirmwareSecurityVersion.rst
```diff
@@ -0,0 +1,30 @@
...
+* Functions
+
+    * ${s}FirmwareGetSecurityVersionExp
+    * ${s}FirmwareSetSecurityVersionExp
+
+* Enumerations
+
+    * ${s}_firmware_security_exp_version_t
+
+Security Version
```
aravindksg, line 29: could you add more literature on objectives for this extension and programming example

scripts/sysman/frmwareSecurityVersion.yml
```diff
@@ -0,0 +1,59 @@
+#
+# Copyright (C) 2023 Intel Corporation
```
aravindksg, line 2: copyright year

scripts/sysman/frmwareSecurityVersion.yml
```diff
@@ -0,0 +1,59 @@
...
+desc: "Anti rollback security version Extension Version(s)"
+version: "1.9"
+name: $s_firmware_security_exp_version_t
+etors:
+    - name: "1_0"
+      value: "$X_MAKE_VERSION( 1, 0 )"
+      desc: "version 1.0"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Get anti roll-back security firmware version number of the current running firmware"
```
aravindksg, line 29: IIRC, the IGSC team mentioned the number here is the version number that would get commited when the user calls "SetSecurityVersion" which is the trigger to write the number right?

also, do we need another out param or another API to provide a "min allowed version number"

scripts/sysman/frmwareSecurityVersion.yml
```diff
@@ -0,0 +1,59 @@
...
+name: SetSecurityVersionExp
+details:
+    - "The application may call this function from simultaneous threads."
+    - "The implementation of this function should be lock-free."
+params:
+    - type: $s_firmware_handle_t
+      name: hFirmware
+      desc: "[in] Handle for the component."
+    - type: char
+      name: "version[$S_STRING_PROPERTY_SIZE]"
```
aravindksg, line 57: setSecurityVersion IIUC really is just a trigger right? I don't think we even need an input version number here?

aravindksg:
@joshuaranjan @saik-intel @AshwinKumarKulkarni

joshuaranjan (commented):
No additional points

scripts/sysman/frmwareSecurityVersion.yml
```diff
@@ -0,0 +1,59 @@
...
+      value: "$X_MAKE_VERSION( 1, 0 )"
+      desc: "version 1.0"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Get anti roll-back security firmware version number of the current running firmware"
+version: "1.9"
+class: $sFirmware
+name: GetSecurityVersionExp
+details:
+    - "The application may call this function from simultaneous threads."
```
saik-intel, line 34: Do we really need Exp API ? As the API is new and its not conflicting or breaking old API.

scripts/sysman/frmwareSecurityVersion.yml
```diff
@@ -0,0 +1,59 @@
...
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Set anti roll-back security firmware version number"
+version: "1.9"
+class: $sFirmware
+name: SetSecurityVersionExp
+details:
+    - "The application may call this function from simultaneous threads."
+    - "The implementation of this function should be lock-free."
+params:
```
saik-intel, line 52: Are we sure that this function implementation should be lock free?  while setting SVN via IGSC, we might need lock right? @aravindksg  @joshuaranjan  @AshwinKumarKulkarni   @bellekal

scripts/sysman/frmwareSecurityVersion.yml
aravindksg: prefer experimental APIs for now, it has better flexibility. we can always promote them in future as needed once APIs are stable

scripts/sysman/frmwareSecurityVersion.yml
bellekal: I think it says **should** and not **must**. From https://spec.oneapi.io/level-zero/latest/core/INTRO.html#terminology **should** means it is recommended and not mandated. So i think it is ok to use "should be lock free".

wdamon-intel (changes requested):
@sanilkumar0 please add narrative/example to the extension programming guide

scripts/sysman/frmwareSecurityVersion.yml
sanilkumar0: GetSecurityVersionExp returns the 'min version allowed', which comes from the fuses

scripts/sysman/frmwareSecurityVersion.yml
sanilkumar0: Good catch.. Yes, we don't need an input here.

wdamon-intel (approved):

PR #284, opened 2024-02-27
Fix typo in handle index for affinity description

matcabral:


matcabral:
corrected the commit message to "typo"

PR #296, opened 2024-03-20
updating versions for scorecard analysis

pbg-intel:
Updated versions for checkout code, runanalysis, upload artifacts, code scanning, enabled publishing to dashboard.


wdamon-intel (approved):

sfblackl-intel (approved):
Looks good on updating versions (which never hurt).  If job fails though, you will need to revert the publish results: true

PR #312, opened 2024-10-14
Fix order by version in DDI tables

bmyates:

### DIFF
--- a/scripts/templates/helper.py
+++ b/scripts/templates/helper.py
@@ -877,7 +877,7 @@ def get_class_function_objs(specs, cname, version = None, includeExt = False):
                     else:
                         objects.append(obj)
 
-    return sorted(objects, key=lambda obj: (float(obj.get('version',"1.0"))*10000) + int(obj.get('ordinal',"100")))
+    return sorted(objects, key=lambda obj: (float(obj.get('version',"1.0").split(".")[1])*10000) + int(obj.get('ordinal',"100")))
 
 """
 Public:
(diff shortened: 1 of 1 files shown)


scripts/templates/helper.py
```diff
@@ -877,7 +877,7 @@ def get_class_function_objs(specs, cname, version = None, includeExt = False):
                     else:
                         objects.append(obj)
 
-    return sorted(objects, key=lambda obj: (float(obj.get('version',"1.0"))*10000) + int(obj.get('ordinal',"100")))
+    return sorted(objects, key=lambda obj: (float(obj.get('version',"1.0").split(".")[1])*10000) + int(obj.get('ordinal',"100")))
```
rwmcguir, line 880: Do we know this address's it?

rwmcguir: What about?
return sorted(objects, key=lambda obj: (float(obj.get('version',"1.0").split(".")[0])*1000000)+(float(obj.get('version',"1.0").split(".")[1])*10000) + int(obj.get('ordinal',"100")))

rwmcguir: FYI.. Did a similar test with this...
exp_objects = [ "2.0", "1.11", "1.9", "1.10", "1.12", "1.8", "1.7" ]
x = sorted(exp_objects, key=lambda obj: ((float(obj.split(".")[0])*10000000) + (float(obj.split(".")[1])*10000)) )

Sadly we have to keep the MAJOR and MINOR in the math sense well away from each other hence the massive difference... Though we still in the future stand to overlap if we ever get to four digit MINORs

bmyates: this fixes the ordering by minor version. Yes it does ignore major version. IMO if/when level zero were to go to 2.0 we would want to substantially rewrite a lot of the driver interface code such as this. 

I'm not opposed to your suggestion though, I will add that in.

rwmcguir: Ya... Python has a number of package utils we can use to properly sort versions.. I was looking at them but they have a flag day in API support around version Python 3.10 (older or newer)... Ideally this concept that the 'ordinal' can override things is broken, was a nice idea, but giving author's ability to break ABI's is not so good. We'd need to lock to a newer version than 3.10 to force something 'natsorted' or 'packageutils'...  seems like to big of a change at the moment.

rwmcguir (approved):

nrspruit (approved):

rwmcguir:
Making large comment here, both Jemale and Neil have now tested this live with L0 CTS and it appears to have fixed the issues..

pbg-intel (approved):
Approved as a hot fix to this specific issue. Other pieces in the script also need patches to handle versioning better.

PR #313, opened 2024-10-14
abi_fix to be verified

pbg-intel:
to verify ddi table abi compatibility issue

### DIFF
--- a/scripts/templates/helper.py
+++ b/scripts/templates/helper.py
@@ -877,7 +877,7 @@ def get_class_function_objs(specs, cname, version = None, includeExt = False):
                     else:
                         objects.append(obj)
 
-    return sorted(objects, key=lambda obj: (float(obj.get('version',"1.0"))*10000) + int(obj.get('ordinal',"100")))
+    return sorted(objects, key=lambda obj: (float(obj.get('version',"1.0").translate({ord('.'):None}))*10000) + int(obj.get('ordinal',"100")))
 
 """
 Public:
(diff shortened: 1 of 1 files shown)


bmyates:
there is already a fix for this issue posted here https://github.com/oneapi-src/level-zero-spec/pull/312

scripts/templates/helper.py
```diff
@@ -895,8 +895,8 @@ def get_class_function_objs_exp(specs, cname):
                     exp_objects.append(obj)
                 else:
                     objects.append(obj)
-    objects = sorted(objects, key=lambda obj: (float(obj.get('version',"1.0"))*10000) + int(obj.get('ordinal',"100")))
-    exp_objects = sorted(exp_objects, key=lambda obj: (float(obj.get('version',"1.0"))*10000) + int(obj.get('ordinal',"100")))              
+    objects = sorted(objects, key=lambda obj: (float(obj.get('version',"1.0").translate({ord('.'):None}))*10000) + int(obj.get('ordinal',"100")))
+    exp_objects = sorted(exp_objects, key=lambda obj: (float(obj.get('version',"1.0").translate({ord('.'):None}))*10000) + int(obj.get('ordinal',"100")))
```
bmyates, line 899: this would translate 2.0 -> 20 which would be less than 1.10 -> 110 but greater than 1.1 -> 11

rwmcguir: If at all possible, I highly recommend avoiding anything to do with the Ordinal, it is a mess waiting to happen...

rwmcguir:
Closed due to fix by another commit: 
Commit e8a03a2

